PR #4, opened 2020-07-15
Add GraphQL API

ddeboer:


src/server/schema.ts
```diff
@@ -0,0 +1,24 @@
...
+    identifier: ID!
+    name: String!
+  }
+
+  type Term {
+    uri: ID!
+    prefLabel: [String]!
+    altLabel: [String]!
+    hiddenLabel: [String]!
+    scopeNote: [String]!
```
ddeboer, line 12: The [current GraphQL API](http://demo.netwerkdigitaalerfgoed.nl:8080/static/graphiql/index.html?query=query%20%7B%0A%20%20__type(name%3A%20%22Term%22)%20%7B%0A%20%20%20%20fields%20%7B%0A%20%20%20%20%20%20name%0A%20%20%20%20%7D%0A%20%20%7D%20%20%20%20%20%0A%7D) also has the fields `definition`, `broader`, `narrower`, `related`.

Do we need these fields if we want to replace the currentAPI?

README.md
```diff
@@ -7,59 +7,114 @@ This experimental application is published for examination and evaluation. It is
...
-npx comunica-sparql \
-    sparql@https://api.data.netwerkdigitaalerfgoed.nl/datasets/rkd/rkdartists/services/rkdartists/sparql \
-    -q "SELECT * WHERE { ?uri schema:name ?label . ?label <bif:contains> 'gogh' } LIMIT 10"
+### Start server
+
+    docker-compose up
+
+### Open the GraphQL editor
+
+http://0.0.0.0:3123/playground
```
ddeboer, line 69: Should this be either `127.0.0.1` or `localhost`?

README.md
```diff
@@ -7,59 +7,114 @@ This experimental application is published for examination and evaluation. It is
...
+
+### List queryable sources
```
ddeboer, line 71: 👍

src/server/schema.ts
sdevalk: Good one. We can do without them for now (the absence of these fields doesn't change the technical working of the prototype), but I'll add these soon.

README.md
sdevalk: Yes, better! `0.0.0.0` works but is somewhat strange - I'll change this. Is this blocking the merge or can I add it later?

ddeboer: Changed.

Issue #6, opened 2020-07-16
Build container image and publish to registry


ddeboer:
@sdevalk currently uses GitLab for this; we’d like to move all CI to GitHub to live with the code. GH caught up with GL’s CI capabilities; for instance, GH now has its own [package registry](https://github.blog/2019-05-10-introducing-github-package-registry/).

Use GitHub Actions to build a container image after each push to master (which we then consider the stable branch) and push that to the [GitHub container registry](https://github.com/features/packages).

PR #8, opened 2020-07-17
Optimize build

ddeboer:
* Ref #6.
* Rename build dir to dist/, because it is more common in Node-land.
* Base builds on Alpine to reduce image size.
* Add production build, reducing image size from 1.05GB to 137MB.
* Use npm ci rather than install to respect package-lock.json.
* Keep dev environment unchanged by adding Dockerfile.dev.
* Simplify readme examples where possible.


README.md
```diff
@@ -11,58 +11,58 @@ This experimental application is published for examination and evaluation. It is
 
 ### Logon to container
 
-    docker-compose run --rm --no-deps --entrypoint /bin/bash node
```
ddeboer, line 14: `--no-deps` is superfluous because there’s only one container defined in `docker-compose.yml`.

README.md
```diff
@@ -11,58 +11,58 @@ This experimental application is published for examination and evaluation. It is
...
+    docker-compose run --rm --entrypoint /bin/sh node
```
ddeboer, line 14: Switching to `sh` because the Alpine Linux distribution (which is smaller) doesn’t ship with `bash`.

README.md
```diff
@@ -11,58 +11,58 @@ This experimental application is published for examination and evaluation. It is
...
 
 ### List queryable sources
 
 This command reads the catalog information in file `configs/catalog.ttl`.
 
-    ./bin/run sources:list
+    bin/run sources:list
```
ddeboer, line 20: Slightly typing or copy/pasting for our readers. 😁

README.md
```diff
@@ -11,58 +11,58 @@ This experimental application is published for examination and evaluation. It is
...
 
-    ./bin/run sources:query --identifiers cht --query fiets > cht.txt
+    bin/run sources:query --identifiers cht --query fiets > cht.txt
 
 ## Query sources via GraphQL
 
 ### Start server
 
-    docker-compose up
+    docker-compose up --build
```
ddeboer, line 65: Readers can start here and still have a clean build.

docker-compose.yml
```diff
@@ -3,11 +3,12 @@ services:
...
     image: network-of-terms-comunica:latest
     user: node
-    build: .
+    build:
+      context: .
+      dockerfile: Dockerfile.dev
     restart: always
     volumes:
-      - .:/usr/src/app
-      - /usr/src/app/node_modules
```
ddeboer, line 10: @sdevalk Not sure why you did this.

package.json
```diff
@@ -39,12 +39,13 @@
   "scripts": {
     "check": "gts check",
     "clean": "gts clean",
-    "compile": "tsc -p .",
+    "build": "tsc",
```
ddeboer, line 42: Seems to be slightly more common than `compile`.

Dockerfile
```diff
@@ -1,9 +1,18 @@
-FROM node:12.0.0
-
-WORKDIR /usr/src/app
-
+FROM node:12.18.2-alpine AS build
```
ddeboer, line 1: Using a so-called multi-stage build here. This stage uses `devDependencies` to build the app, while the stage below (line 8 ff.) will be smaller because it only has JS, not TS files and dependencies.

package.json
```diff
@@ -39,12 +39,13 @@
...
     "fix": "gts fix",
     "posttest": "npm run check",
+    "start": "node dist/server/server.js",
```
ddeboer, line 45: In the production stage (see `Dockerfile`) we have only JS (not TS), so start the server using JS instead of `ts-node-dev`.

package.json
sdevalk: I'm not at all religious about this :) but I've used Google's `gts` configurator here in order to have some agreement on naming things. `gts` uses `compile` but also `build` instead of `dist` (https://github.com/google/gts).

ddeboer: Good point; reverted my change from `dist` to `build`.

docker-compose.yml
sdevalk: This 'trick' enables fast development: using this, a change on the host (e.g. because I update a source file) is instantly reflected in the container. See [here](https://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html#the-node_modules-volume-trick) for example. However, I'm open to improvement!

Dockerfile.dev
```diff
@@ -0,0 +1,5 @@
+FROM node:12.18.2-alpine
```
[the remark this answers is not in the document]
ddeboer: Nice! Using a single Dockerfile now.

docker-compose.yml
ddeboer: Ah yes, the `node_modules/` volume makes sure that the `node_modules/` directory that is filled during `docker-compose build` is still available after starting the containers with `docker-compose up`. So reverted.

PR #9, opened 2020-07-17
Publish Docker image to GitHub Packages

### LINKED ISSUE #6: Build container image and publish to registry

ddeboer:
Fix #6.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,24 @@
+name: Build app and push Docker image
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Build and push image to GitHub Packages
+        uses: docker/build-push-action@v1.1.0
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -0,0 +1,24 @@
+name: Build app and push Docker image
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
```
ddeboer, line 8: We’re not using tags yet, but just in case.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,24 @@
+name: Build app and push Docker image
+
+on:
+  push:
+    branches:
+      - master
```
ddeboer, line 6: We should (at least in theory) consider our master branch as stable and deployable, so I build the package from that.

PR #15, opened 2020-08-06
Upgrade to gts version 2 (including eslint instead of tslint)

sdevalk:


package.json
```diff
@@ -14,9 +14,12 @@
   "dependencies": {
     "@comunica/actor-init-sparql": "1.x.x",
     "@comunica/actor-init-sparql-rdfjs": "1.x.x",
+    "@comunica/bus-query-operation": "^1.13.0",
+    "@comunica/core": "^1.13.0",
```
ddeboer, line 18: This is not strictly necessary: upgrading only gts to its version 2 can be done by running

```
npm i gts@latest
```

in the current master branch.

package.json
```diff
@@ -14,9 +14,12 @@
...
     "@hapi/hoek": "9.x.x",
     "@hapi/joi": "17.x.x",
     "@oclif/command": "1.x.x",
+    "@rdfjs/data-model": "^1.1.2",
```
ddeboer, line 22: Is this related to the gts upgrade?

package.json
sdevalk: I've not only upgraded gts, I've also applied its linting rules. One of those rules states that 'extraneous packages' are forbidden, i.e. packages that are used in the code but that are not declared in package.json. This is the reason why I've added `@comunica/bus-query-operation`, `@comunica/core` and `@rdfjs/data-model` here.

package.json
sdevalk: Yes, it does - `@rdfjs/data-model` is used in the code but wasn't declared here, which gts forbids.

PR #16, opened 2020-08-06
Use catalog from network-of-terms-catalog

### LINKED ISSUE #5: Cache catalog and queries on server start
Currently, the `catalog.ttl` and `.rq` query files are read on each request. We can optimise this by reading and caching them on server start.

ddeboer:
Fix #5.


src/commands/sources/list.ts
```diff
@@ -1,7 +1,9 @@
-import {CatalogService, Distribution} from '../../services/catalog';
 import {cli} from 'cli-ux';
 import {Command, flags} from '@oclif/command';
-import * as Logger from '../../helpers/logger';
+import {
+  Catalog,
+  Dataset,
+} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
 
 export class ListSourcesCommand extends Command {
```
sdevalk, line 8: The `flags` aren't used anymore. Remove these?

src/services/catalog.ts
```diff
@@ -1,167 +0,0 @@
-import {ActorInitSparql} from '@comunica/actor-init-sparql/lib/ActorInitSparql-browser';
```
sdevalk, line 1: This file - `catalog.ts` - still exists, though empty. Remove it?

src/services/distributions.ts
```diff
@@ -1,71 +1,70 @@
...
-    const accessService = await this.catalogService.getAccessServiceByDistributionId(
-      args.distributionId
-    );
-    if (accessService === null) {
+    this.logger.info(`Preparing to query source "${args.source}"...`);
+    const dataset = await this.catalog.getByIdentifier(args.source);
+    if (dataset === undefined) {
       throw Error(
-        `Access service of source "${args.distributionId}" not found in catalog`
+        `Dataset with identifier "${args.identifier}" not found in catalog`
```
sdevalk, line 52: This throws the error underneath if a non-existing source is given:

`Dataset with identifier "undefined" not found in catalog`

Also, perhaps change the word `Dataset` to `Source`?

src/services/query.ts
```diff
@@ -65,29 +65,32 @@ export class QueryTermsService {
...
-      this.accessService.query,
+      this.dataset.distribution.query,
       config
     )) as IActorQueryOperationOutputQuads;
 
     return new Promise((resolve, reject) => {
       const termsTransformer = new TermsTransformer();
-      result.quadStream.on('error', reject);
+      result.quadStream.on('error', error => {
+        console.log('error', error);
```
sdevalk, line 80: Is it necessary to log the error here? Can we leave it to the caller and simply reject instead?

src/services/terms.ts
```diff
@@ -32,7 +32,7 @@ export class TermsTransformer {
 
   fromQuad(quad: RDF.Quad): void {
     const subject = quad.subject;
-    if (!subject.equals(this.currentTerm.id)) {
+    if (this.currentTerm.id === undefined || !subject.equals(this.currentTerm.id)) {
```
sdevalk, line 35: The new condition isn't really necessary: if `this.currentTerm.id` is undefined, `subject.equals()` returns true?

ddeboer: This caused `gts check` violation before.

ddeboer: Just

```
if (!subject.equals(this.currentTerm.id)) {
```

gives:

```
Error:(35, 25) TS2345: Argument of type 'Term | undefined' is not assignable to parameter of type 'Term'.
  Type 'undefined' is not assignable to type 'Term'.
```

ddeboer:
Thanks for your feedback! Most issues are resolved now.

src/services/terms.ts
sdevalk: Good one! :)

(Somewhat strange, though - the [RDFJS spec](https://rdf.js.org/data-model-spec/#term-interface) states that `equals()` returns `true` if (among others) the parameter is not `undefined`, so I expected `undefined` to be a valid value.)

PR #23, opened 2020-08-21
Share single Comunica instance

ddeboer:
Fix netwerk-digitaal-erfgoed/bruikbaar#31.

### DIFF
--- a/src/commands/sources/query.ts
+++ b/src/commands/sources/query.ts
@@ -6,6 +6,7 @@ import {QueryResult} from '../../services/query';
 import * as RDF from 'rdf-js';
 import {Term} from '../../services/terms';
 import {Catalog} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
+import {newEngine} from '@comunica/actor-init-sparql';
 
 interface Row {
   distributionTitle: string;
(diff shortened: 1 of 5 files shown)


src/commands/sources/query.ts
```diff
@@ -81,7 +82,8 @@ export class QuerySourcesCommand extends Command {
       level: flags.loglevel,
     });
     const catalog = await Catalog.default();
-    const service = new DistributionsService({logger, catalog});
+    const comunica = newEngine();
```
sdevalk, line 85: Looks good! We may need to change this at some point, though: this Comunica instance is bound to SPARQL endpoints. If we want to query other source types (e.g. custom Web APIs) we'll have to use different Comunica instances.

src/services/distributions.ts
```diff
@@ -2,15 +2,18 @@ import * as Joi from '@hapi/joi';
 import Pino from 'pino';
 import {QueryResult, QueryTermsService} from './query';
 import {Catalog} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
+import {IActorInitSparqlArgs} from '@comunica/actor-init-sparql/lib/ActorInitSparql-browser';
```
sdevalk, line 5: To be sure: is this - `IActorInitSparqlArgs` - the correct type? File `src/services/query.ts` uses `ActorInitSparql`.

ddeboer: “Depend on abstractions, not on concretions”, so I used the interface rather than the class here:

```
export declare class ActorInitSparql extends ActorInit implements IActorInitSparqlArgs {
```

I do agree this interface name is quite confusing, though.

sdevalk: 👍🏻

PR #24, opened 2020-08-25
Upgrade network-of-terms-catalog to 2.0.0

### LINKED ISSUE #17: Identify datasets by their IRI rather than identifier
The dataset identifiers that we now use are made-up (albeit common) abbreviations of dataset names, for instance ‘rkdartists’ and ‘cht’. Should we replace these identifiers with proper IRIs?

Some considerations:

- Short strings like our current identifiers look more familiar to GraphQL users than IRIs.
- When the number of term sources grows, identifier collisions become more likely, especially if they are provided by term source managers themselves rather than NDE.
- IRIs can be resolved for further information about the dataset, whereas identifiers are a dead end.

@sdevalk Should we allow identification _only_ by IRI or also keep supporting the identifiers?

ddeboer:
* Fix #17.
* Adapt to https://github.com/netwerk-digitaal-erfgoed/network-of-terms-catalog/pull/9.
* Replace dataset identifier with distribution IRI in both code and docs.

### DIFF
--- a/src/commands/sources/list.ts
+++ b/src/commands/sources/list.ts
@@ -21,7 +21,7 @@ export class ListSourcesCommand extends Command {
         get: (distribution: DatasetDistribution) => distribution.dataset.name,
       },
       distributionId: {
-        header: 'Distribution URI',
+        header: 'Source URI',
         get: (distribution: DatasetDistribution) =>
           distribution.distribution.iri.toString(),
       },
--- a/src/commands/sources/list.ts
+++ b/src/commands/sources/list.ts
@@ -3,26 +3,40 @@ import {Command} from '@oclif/command';
 import {
   Catalog,
   Dataset,
+  Distribution,
 } from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
+
+type DatasetDistribution = {
+  dataset: Dataset;
+  distribution: Distribution;
+};
 
 export class ListSourcesCommand extends Command {
   static description = 'List queryable sources';
 
-  protected render(datasets: readonly Dataset[]): void {
-    cli.table(datasets as Dataset[], {
+  protected render(distributions: readonly DatasetDistribution[]): void {
+    cli.table(distributions as DatasetDistribution[], {
       distributionTitle: {
         header: 'Source Name',
-        get: (dataset: Dataset) => dataset.name,
+        get: (distribution: DatasetDistribution) => distribution.dataset.name,
       },
       distributionId: {
-        header: 'Source ID',
-        get: (dataset: Dataset) => dataset.identifier,
+        header: 'Distribution URI',
+        get: (distribution: DatasetDistribution) =>
+          distribution.distribution.iri.toString(),
       },
     });
   }
 
   async run(): Promise<void> {
     const catalog = await Catalog.default();
-    this.render(catalog.datasets);
+    const distributions = catalog.datasets.flatMap((dataset: Dataset) =>
+      dataset.distributions.map(distribution => ({
+        dataset,
+        distribution,
+      }))
+    );
+
+    this.render(distributions);
   }
 }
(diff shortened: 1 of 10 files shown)


tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./node_modules/gts/tsconfig-google.json",
   "compilerOptions": {
+    "lib": ["es2019"],
```
ddeboer, line 4: Necessary for `Array.flatMap()`.

src/commands/sources/query.ts
```diff
@@ -36,12 +36,14 @@ export class QuerySourcesCommand extends Command {
...
-  protected render(results: QueryResult[]): void {
+  protected render(results: QueryResult[], catalog: Catalog): void {
     const rowsPerDistribution = results.map((result: QueryResult): Row[] => {
       return result.terms.map(
         (term: Term): Row => {
           return {
-            distributionTitle: result.dataset.distribution.url.toString(),
+            datasetTitle:
+              catalog.getDatasetByDistributionIri(result.distribution.iri)
+                ?.name ?? '',
```
ddeboer, line 46: Super ugly, but this is what `gts fix` makes of it.

src/commands/sources/query.ts
```diff
@@ -19,9 +19,9 @@ export class QuerySourcesCommand extends Command {
   static description = 'Query sources for terms';
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   static flags: flags.Input<any> = {
-    identifiers: flags.string({
+    distributions: flags.string({
       description:
-        'Identifiers of sources to query, comma-separated, e.g. "nta,rkdartists"',
+        'URIs of distributions to query, comma-separated, e.g. "https://www.wikidata.org/sparql,https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql"',
```
ddeboer, line 24: Though we’re using ‘IRI’ internally, I decided to keep ‘URI’ for all public-facing bits, including the GraphQL schema and this here.

src/commands/sources/list.ts
```diff
@@ -3,26 +3,40 @@ import {Command} from '@oclif/command';
...
+    cli.table(distributions as DatasetDistribution[], {
       distributionTitle: {
         header: 'Source Name',
-        get: (dataset: Dataset) => dataset.name,
+        get: (distribution: DatasetDistribution) => distribution.dataset.name,
       },
       distributionId: {
-        header: 'Source ID',
-        get: (dataset: Dataset) => dataset.identifier,
+        header: 'Distribution URI',
```
sdevalk, line 24: Or `Source URI` rather than `Distribution URI` (so as to hide that we use distributions internally)?

src/commands/sources/query.ts
sdevalk, line 24: Or use `URIs of sources` instead?

src/commands/sources/query.ts
sdevalk: 😢 😄

src/services/distributions.ts
```diff
@@ -51,25 +51,25 @@ export class DistributionsService {
   async query(options: QueryOptions): Promise<QueryResult> {
     const args = Joi.attempt(options, schemaQuery);
     this.logger.info(`Preparing to query source "${args.source}"...`);
-    const dataset = await this.catalog.getByIdentifier(args.source);
+    const dataset = await this.catalog.getDatasetByDistributionIri(args.source);
     if (dataset === undefined) {
       throw Error(
-        `Source with identifier "${args.source}" not found in catalog`
+        `Source with distribution IRI "${args.source}" not found in catalog`
```
sdevalk, line 57: This error may bubble up to a user (e.g. via the CLI). Perhaps avoid the word 'distribution' (and 'catalog' and 'IRI'?) and merely use `Source with URI xxx not found` (words that our interfaces [CLI, GraphQL] use)?

sdevalk (approved):
👍🏻

PR #25, opened 2020-09-01
Deploy application

ddeboer:
Deploy the application to a managed Kubernetes cluster on DigitalOcean.


ddeboer:
Until I get access to the NDE Digital Ocean account, this is temporarily deployed on my own account at https://termen.ddeboer.nl/playground.

sdevalk (commented):
Looks good! One question: would it be possible to keep the file outside of this package? The file is specific to our prototype instance of the NoT. Ideally the package should be usable by everyone for creating and deploying their own instance. It would then help if the package doesn't contain instance-specific files.

ddeboer:
Which file, exactly? I already removed the Kubernetes manifest itself from here. It’s quite natural to have GitHub workflows checked into the repository. Excluded that directory from the Docker image, so when people pull that they won’t be confused.

sdevalk (approved):

PR #26, opened 2020-09-09
Add dataset creators to GraphQL endpoint

ddeboer:


src/server/schema.ts
```diff
@@ -2,6 +2,12 @@ export const schema = `
   type Source {
     uri: ID!
     name: String!
+    creators: [Creator]!
```
ddeboer, line 5: In practice, we only have a single creator for each source in the catalog. Should we be open for extension by having a list here or simplify and keep it a single value?

sdevalk (approved):
👍🏻

src/server/schema.ts
sdevalk: Good question. There are some sources created by more than one organization. But in practice they use the name of their partnership, not their individual names. However, I think we should stick to a flexible data structure in our API and allow for multiple creators.

PR #30, opened 2020-09-14
Return related terms

ddeboer:
* Upgrade to netwerk-digitaal-erfgoed/network-of-terms-catalog#17.
* Deduplicate prefLabels.
* Add examples to readme.


src/services/terms.ts
```diff
@@ -1,20 +1,38 @@
...
+  broader: RDF.Term[] = [];
+  narrower: RDF.Term[] = [];
 }
 
 export class TermsTransformer {
-  protected terms: Term[] = [];
-  protected currentTerm: Term = new Term();
-  protected readonly predicateToPropertyMap = new Map<string, string>([
+  private termsIris: string[] = [];
+  private termsMap: Map<string, SparqlResultTerm> = new Map();
```
ddeboer, line 34: I had to replace the simple `currentTerm` state management with:

- a list that is ordered and tracks all search results (i.e. top-level terms in the SPARQL result)
- a map for quick lookups of related terms.

src/services/terms.ts
```diff
@@ -1,20 +1,37 @@
...
 export class Term {
-  id: RDF.Term | undefined = undefined;
+  constructor(
+    readonly id: RDF.Term,
+    readonly type: RDF.Term | undefined,
+    readonly prefLabels: RDF.Term[],
+    readonly altLabels: RDF.Term[],
+    readonly hiddenLabels: RDF.Term[],
+    readonly scopeNotes: RDF.Term[],
+    readonly broader: RelatedTerm[],
```
sdevalk, line 11: 'prefLabels', 'altLabels', etc, are in plural, but 'broader' and 'narrower' are not. Does it make sense to rename these to e.g. 'broaderTerms' and 'narrowerTerms'?

src/services/terms.ts
sdevalk: 👍🏻

sdevalk (commented):
Nice!

src/services/terms.ts
ddeboer: It sure does! As we have simple, singular words in the GraphQL schema (`altLabel` etc.) I decided to keep `broader`/`narrower` there.

PR #31, opened 2020-09-17
Redirect / to /playground

ddeboer:
Currently, / shows a 404, so redirect to /playground to give
users a meaningful homepage.


sdevalk (approved):
Good one!

PR #32, opened 2020-09-24
Migrate to GitHub Container Registry (GHCR)

ddeboer:
* GHCR is replacing GitHub Packages for Docker images.
* This enables us to make the image public.
* Rename Docker image to network-of-terms-api.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -16,17 +16,28 @@ jobs:
 
       - name: Set variables
         id: vars
-        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+        run: |
+          echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+          echo "::set-output name=docker_image::ghcr.io/netwerk-digitaal-erfgoed/network-of-terms-api"
 
-      - name: Build and push image to GitHub Packages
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -16,17 +16,28 @@ jobs:
 
       - name: Set variables
         id: vars
-        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+        run: |
+          echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+          echo "::set-output name=docker_image::ghcr.io/netwerk-digitaal-erfgoed/network-of-terms-api"
```
ddeboer, line 21: TBD: is `network-of-terms-api` a good name? At the very least, we should drop the `comunica` because that’s not relevant to people using the image.

Alternatives: `network-of-terms-server`, `network-of-terms-graphql-server`, …?

ddeboer: > Rename the repo `network-of-terms-comunica` to `network-of-terms-api`?

Yes, done!

> For example: `network-of-terms-browser-comunica` and `network-of-terms-browser-graphql`?

Sounds good to me, as on the browser-side the client _would_ have to make a choice between Comunica or GraphQL.

PR #35, opened 2020-10-03
fix: resolve source errors

### LINKED ISSUE #34: Make queries fault-tolerant
When a single source fails, it causes all results to fail. Probably related to our use of `Promise.all()`.

For instance:

```graphql
query Terms {
  terms(sources: ["http://vocab.getty.edu/aat/sparql", "https://data.netwerkdigitaalerfgoed.nl/NMVW/thesaurus/sparql"], query: "griek*") {
    source {
      uri
      name
      creators {
        uri
        name
        alternateName
      }
    }
    terms {
      uri
      prefLabel
      altLabel
      hiddenLabel
      scopeNote
    }
  }
}
```

sdevalk:
Fix #34.

@ddeboer This is a first shot at [34](https://github.com/netwerk-digitaal-erfgoed/network-of-terms-api/issues/34) and open to debate.

Some background:

1. The issue is indeed caused by `Promise.all()`.
2. We could fix it with `Promise.allSettled()`, i.e. by removing the source that failed and continuing with the sources that succeeded. However, this way the user doesn't get to see the failed source in the response (e.g. GraphQL output), which is somewhat strange: if I query two sources and one of these fails, I still would like to see that source in the response (otherwise it suggests that the failed source wasn't queried at all).
3. The PR ensures that all sources resolve, even if one of these failed (in which case an empty result set is returned). We could improve this by also returning the error of the source (e.g. as part of the `QueryResult` interface in [query.ts](https://github.com/netwerk-digitaal-erfgoed/network-of-terms-api/blob/master/src/services/query.ts#L33)).

### DIFF
--- a/src/services/query.ts
+++ b/src/services/query.ts
@@ -78,7 +78,7 @@ export class QueryTermsService {
       config
     )) as IActorQueryOperationOutputQuads;
 
-    return new Promise((resolve /*, reject */) => {
+    return new Promise(resolve => {
       const termsTransformer = new TermsTransformer();
       result.quadStream.on('error', (error: Error) => {
         this.logger.error(
--- a/src/services/query.ts
+++ b/src/services/query.ts
@@ -80,7 +80,12 @@ export class QueryTermsService {
 
     return new Promise((resolve, reject) => {
       const termsTransformer = new TermsTransformer();
-      result.quadStream.on('error', reject);
+      result.quadStream.on('error', (error: Error) => {
+        this.logger.error(
+          `An error occurred when querying "${this.distribution.endpoint}": ${error}`
+        );
+        resolve({distribution: this.distribution, terms: []});
+      });
       result.quadStream.on('data', (quad: RDF.Quad) =>
         termsTransformer.fromQuad(quad)
       );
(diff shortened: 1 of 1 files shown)


src/services/query.ts
```diff
@@ -78,9 +78,14 @@ export class QueryTermsService {
       config
     )) as IActorQueryOperationOutputQuads;
 
-    return new Promise((resolve, reject) => {
+    return new Promise((resolve /*, reject */) => {
```
ddeboer, line 81: You can just remove the `reject` argument.

ddeboer:
> 3. The PR ensures that all sources resolve, even if one of these failed (in which case an empty result set is returned). We could improve this by also returning the error of the source (e.g. as part of the QueryResult interface in query.ts).

Agreed; let’s extend the `Terms` GraphQL object with an `errors` list property. Unfortunately, Comunica returns really unhelpful error messages, but we start with just having a default error message ‘query failed’ or something similar. Would you like to add this in a separate PR?

sdevalk:
> > 1. The PR ensures that all sources resolve, even if one of these failed (in which case an empty result set is returned). We could improve this by also returning the error of the source (e.g. as part of the QueryResult interface in query.ts).
> 
> Agreed; let’s extend the `Terms` GraphQL object with an `errors` list property. Unfortunately, Comunica returns rather unhelpful error messages, but we can start with just having a default error message ‘query failed’ or something similar. Would you like to add this in a separate PR?

Will do! :)

PR #55, opened 2020-10-17
feat: Return errors

### LINKED ISSUE #52: Return errors when source is unreachable
At least a certain set of errors are scoped to the sources we query, which makes GraphQL’s top-level `error` property less suitable. Therefore, add types to our schema, introducing a union type `Terms | TimeoutError | ServerError` (or something similar). Clients can then conditionally retrieve results.

For inspiration: https://medium.com/@sachee/200-ok-error-handling-in-graphql-7ec869aec9bc and https://blog.logrocket.com/handling-graphql-errors-like-a-champ-with-unions-and-interfaces/.

Try to keep the query interface as simple as possible and not break BC.

ddeboer:
* Fix #52.
* Add two error classes: `ServerError` (which we already use)
  and `TimeoutError` (which we can start using when we support
  timeouts).
* Deprecate `terms` but don’t remove it yet to prevent BC breaks.
  Recommend `result` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ query Terms {
     }
     result {
       __typename 
-      ... on Success {
+      ... on Terms {
         terms {
           uri
           prefLabel
@@ -179,7 +179,7 @@ query Terms {
     }
     result {
       __typename
-      ... on Success {
+      ... on Terms {
         terms {
           uri
           prefLabel
--- a/README.md
+++ b/README.md
@@ -133,23 +133,31 @@ query Terms {
         alternateName
       }
     }
-    terms {
-      uri
-      prefLabel
-      altLabel
-      hiddenLabel
-      scopeNote
-      broader {
-        uri
-        prefLabel
+    result {
+      __typename 
+      ... on Success {
+        terms {
+          uri
+          prefLabel
+          altLabel
+          hiddenLabel
+          scopeNote
+          broader {
+            uri
+            prefLabel
+          }
+          narrower {
+            uri
+            prefLabel
+          }
+          related {
+            uri
+            prefLabel
+          }
+        }
       }
-      narrower {
-        uri
-        prefLabel
-      }
-      related {
-        uri
-        prefLabel
+      ... on Error {
+        message
       }
     }
   }
@@ -169,12 +177,20 @@ query Terms {
         alternateName
       }
     }
-    terms {
-      uri
-      prefLabel
-      altLabel
-      hiddenLabel
-      scopeNote
+    result {
+      __typename
+      ... on Success {
+        terms {
+          uri
+          prefLabel
+          altLabel
+          hiddenLabel
+          scopeNote
+        }
+      }
+      ... on Error {
+        message
+      }
     }
   }
 }
--- a/src/server/schema.ts
+++ b/src/server/schema.ts
@@ -30,11 +30,30 @@ export const schema = `
 
   type Terms {
     source: Source!
-    terms: [Term]!
+    terms: [Term]! @deprecated(reason: "Use 'result' instead")
+    result: Result!
   }
 
   type Query {
     terms(sources: [ID]!, query: String!): [Terms]
     sources: [Source]
+  }
+  
+  union Result = Success | TimeoutError | ServerError
+  
+  type Success {
+    terms: [Term]
+  }
+  
+  type TimeoutError implements Error {
+    message: String!
+  }
+  
+  type ServerError implements Error {
+    message: String!
+  }
+  
+  interface Error {
+    message: String!
   }
 `;
--- a/src/server/schema.ts
+++ b/src/server/schema.ts
@@ -28,20 +28,20 @@ export const schema = `
     prefLabel: [String]!
   }
 
-  type Terms {
+  type TermsQueryResult {
     source: Source!
     terms: [Term]! @deprecated(reason: "Use 'result' instead")
-    result: Result!
+    result: TermsResult!
   }
 
   type Query {
-    terms(sources: [ID]!, query: String!): [Terms]
+    terms(sources: [ID]!, query: String!): [TermsQueryResult]
     sources: [Source]
   }
   
-  union Result = Success | TimeoutError | ServerError
+  union TermsResult = Terms | TimeoutError | ServerError
   
-  type Success {
+  type Terms {
     terms: [Term]
   }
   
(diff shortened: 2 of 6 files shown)


ddeboer:
The `Error.message` is the low-level error message from Comunica, the N3 parser etc. That’s not super user-friendly, but it can help us debug issues. End users could just look at `__typename`: `ServerError` and/or `TimeoutError`.

```
"result": {
  "__typename": "ServerError",
  "message": "Unexpected graph on line 1."
},
```

README.md
```diff
@@ -133,23 +133,31 @@ query Terms {
...
-      hiddenLabel
-      scopeNote
-      broader {
-        uri
-        prefLabel
-      }
-      narrower {
-        uri
-        prefLabel
+    result {
```
sdevalk, line 136: Thinking out load: is it useful to make the `source` object also a part of the `result` structure? With our current implementation it's unlikely that errors will occur when loading the `source` information, but would it make the API design more forward-compatible?

src/services/query.ts
```diff
@@ -30,10 +30,23 @@ const schemaConstructor = Joi.object({
   comunica: Joi.object().required(),
 });
 
-export interface QueryResult {
-  distribution: Distribution;
-  terms: Term[];
+export type QueryResult = Success | TimeoutError | ServerError;
```
sdevalk, line 33: These type and class names make perfect sense, but also are somewhat difficult to distinguish at first sight. For example: what's the difference between a `QueryResult` and just a `Result`? And why is the query error named `QueryError` but the query success just `Success`?

sdevalk (changes requested):
Nice work! :) Just some minor questions.

README.md
ddeboer: The errors I added here I scoped per source, so I’d prefer to keep the source data (always available because it comes from our catalog) out of the variable bit (success with terms or failure with error message).

I case of total breakage (which obviously should occur as little as possible 😅), we can still use the GraphQL `errors` property on the global level.

src/services/query.ts
ddeboer: Good point! The reason is a bad one: I already added a `Result`, so this had to be named differently and became `QueryResult`. I simplified the type names; are they clearer now?

src/server/schema.ts
```diff
@@ -30,11 +30,30 @@ export const schema = `
 
   type Terms {
     source: Source!
-    terms: [Term]!
+    terms: [Term]! @deprecated(reason: "Use 'result' instead")
```
ddeboer, line 33: Kept this for BC, but deprecated it to point clients to the new `result` property.

README.md
sdevalk: 👍🏻

src/services/query.ts
sdevalk: Perfect, thanks!

src/server/schema.ts
sdevalk: 👍🏻

ddeboer: By the way, an alternative approach would be to keep `terms` (and don’t deprecate it) and _add_ an `errors` list property (that can be empty) instead of using a union type.

ddeboer:
By the way, an alternative approach would be to keep terms (and don’t deprecate it) and add an errors list property (that can be empty) instead of using a union type.

sdevalk (approved):
👍🏻

sdevalk:
> By the way, an alternative approach would be to keep `terms` (and don’t deprecate it) and _add_ an `errors` list property (that can be empty) instead of using a union type.

Does this mean that we don't need the `... on Success` and `... on Error` constructs? For example:

```graphql
terms(sources: [...], query: "...") {
  terms {
    uri
    prefLabel
    altLabel
    hiddenLabel
    scopeNote
  }
  error {
    message
  }
}
```

ddeboer:
Exactly, so it’s a bit easier for clients. But keeping the errors separate has a downside as well: it’s less explicit how `term` and `errors` are related. After all, `terms` _or_ `errors` apply (if you have one, you won’t have the other), but we have to include both properties. Moreover, clients may forget to retrieve the `errors` property and miss the reason for an empty list of terms. The union forces clients to handle both the success and the error state.

As you can probably tell, I’m still on the fence as to what’s the best way. If you like, we can discuss this on Thursday and leave this PR open until then.

sdevalk:
> Exactly, so it’s a bit easier for clients. But keeping the errors separate has a downside as well: it’s less explicit how `term` and `errors` are related. After all, `terms` _or_ `errors` apply (if you have one, you won’t have the other), but we have to include both properties, which obscures their exclusivity. Moreover, clients may forget to retrieve the `errors` property and miss the reason for an empty list of terms. The union type forces clients to handle both the success and the error state.
> 
> As you can probably tell, I’m still on the fence as to what’s the best way. If you like, we can discuss this on Thursday and leave this PR open until then.

Thursday it is! 🤓

ddeboer:
We decided to rethink the GraphQL queries including (future) concerns about errors related to fetching source data.

Until then, let’s mark this PR as draft.

src/server/schema.ts
```diff
@@ -30,11 +30,30 @@ export const schema = `
...
+    result: Result!
   }
 
   type Query {
     terms(sources: [ID]!, query: String!): [Terms]
     sources: [Source]
   }
+  
+  union Result = Success | TimeoutError | ServerError
```
ddeboer, line 42: It’s not possible to say `union Result = [Term] | TimeoutError | ServerError` here.

@sdevalk Do you then still want to change the name from `Success` to `Terms`? That would give:

```graphql
... on Terms {
  terms {
    uri
    prefLabel
  }
}
```

sdevalk: Good one. `... on Terms` + `terms` sounds a bit redundant.

`... on Success` and `... on Error` are nice counterparts. Would that be an option, after all? Or should we use a unique name here, e.g. `... on TermsSuccess`, to prevent conflicts with a (future) `... on Success` in the `source` section (`... on SourceSuccess`)?

ddeboer: > `... on Terms` + `terms` sounds a bit redundant.

Agreed. Still, it reads okay: on receiving some terms (`... on Terms`) rather than an error (`... on Error`), unpack those terms and retrieve properties from them. 

So I decided to keep rename the type `Result` (too generic as we will have _source_ results in the future, too) to `Terms`.

sdevalk (approved):
💪

PR #75, opened 2020-11-06
feat: Sort results alphabetically by prefLabel

### LINKED ISSUE #68: Sort results alphabetically
This is to prevent questions about the order that search results appear in. Our SPARQL `CONSTRUCT` queries preclude us from keeping the source’s ordering or providing our own `ORDER BY`. The sort should therefore happen on the client-side (resolvers?).

ddeboer:
Fix #68.

### DIFF
--- a/src/services/query.ts
+++ b/src/services/query.ts
@@ -95,7 +95,9 @@ export class QueryTermsService {
         termsTransformer.fromQuad(quad)
       );
       result.quadStream.on('end', () => {
-        const terms = termsTransformer.asArray();
+        const terms = termsTransformer
+          .asArray()
+          .sort(alphabeticallyByPrefLabel);
         this.logger.info(
           `Found ${terms.length} terms matching "${this.query}" in "${
             this.distribution.endpoint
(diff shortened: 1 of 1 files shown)


src/services/query.ts
```diff
@@ -106,3 +108,17 @@ export class QueryTermsService {
...
+  const prefLabelB = b.prefLabels[0]?.value;
+  if (prefLabelA < prefLabelB) {
+    return -1;
+  }
+
+  if (prefLabelA > prefLabelB) {
+    return 1;
+  }
+
+  return 0;
```
sdevalk, line 123: Or we can use `return prefLabelA.localeCompare(prefLabelB)` - a bit more concise?

sdevalk (changes requested):

src/services/query.ts
ddeboer: Nice! Changed.

src/services/query.ts
```diff
@@ -106,3 +108,9 @@ export class QueryTermsService {
     });
   }
 }
+
+const alphabeticallyByPrefLabel = (a: Term, b: Term) => {
+  const prefLabelA = a.prefLabels[0]?.value ?? '';
+  const prefLabelB = b.prefLabels[0]?.value ?? '';
```
ddeboer, line 114: `localCompare` must take a non-null string, so default to an empty string here.

sdevalk: 👍🏻

sdevalk (approved):

PR #146, opened 2021-04-17
Feat: extended sorting

EnnoMeijers:
Extended the sorting of the results to include the altLabels. This is helpful when many prefLabels have the same value f.g. searching for 'markt' in Wikidata streets.

### DIFF
--- a/src/services/query.ts
+++ b/src/services/query.ts
@@ -95,9 +95,7 @@ export class QueryTermsService {
         termsTransformer.fromQuad(quad)
       );
       result.quadStream.on('end', () => {
-        const terms = termsTransformer
-          .asArray()
-          .sort(alphabeticallyByPrefLabel);
+        const terms = termsTransformer.asArray().sort(alphabeticallyByLabel);
         this.logger.info(
           `Found ${terms.length} terms matching "${this.query}" in "${
             this.distribution.endpoint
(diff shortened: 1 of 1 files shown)


src/services/query.ts
```diff
@@ -109,8 +107,12 @@ export class QueryTermsService {
   }
 }
 
-const alphabeticallyByPrefLabel = (a: Term, b: Term) => {
+const alphabeticallyByLabel = (a: Term, b: Term) => {
```
sdevalk, line 110: Or: `alphabeticallyByLabels` (in plural), since it sorts on both a prefLabel and altLabel?

src/services/query.ts
```diff
@@ -109,8 +107,12 @@ export class QueryTermsService {
...
   const prefLabelA = a.prefLabels[0]?.value ?? '';
+  const altLabelA = a.altLabels[0]?.value ?? '';
+  const sortLabelA = prefLabelA.concat(altLabelA);
```
sdevalk, line 113: I think in JavaScript it's a bit more common (and slightly faster) to use the string concat operator rather than  `concat()`, e.g. `const sortLabelA = prefLabelA + altLabelA;`

src/services/query.ts
```diff
@@ -109,8 +107,12 @@ export class QueryTermsService {
   }
 }
 
-const alphabeticallyByPrefLabel = (a: Term, b: Term) => {
+const alphabeticallyByLabel = (a: Term, b: Term) => {
   const prefLabelA = a.prefLabels[0]?.value ?? '';
+  const altLabelA = a.altLabels[0]?.value ?? '';
```
sdevalk, line 112: A term could have multiple altLabels. Is it relevant to take all labels into account, not just the first one?

sdevalk (commented):
Nice improvement!

src/services/query.ts
EnnoMeijers: Goed idee!

src/services/query.ts
EnnoMeijers: I was on this track before try to operate it on the conditional statement directly but that didn't work. This is indeed a clearer approach.

src/services/query.ts
EnnoMeijers: I haven't found a use case that demonstated the need for sorting on all the altLabels. My suggestion is to start with just the first altLabel and see if we run into this use case at some point.

Issue #36, opened 2020-10-03
SPARQL query limit haphazardly breaks result sets


ddeboer:
[...]

PR #151, opened 2021-04-26
fix: Guard against incomplete related terms

### LINKED ISSUE #143: Error: "Cannot read property 'id' of undefined"
If I send the query underneath to the Network of Terms...

https://termennetwerk-api.netwerkdigitaalerfgoed.nl/playground

```graphql

query Terms {
  terms(
    sources: [
      "https://data.cultureelerfgoed.nl/PoolParty/sparql/term/id/cht"
    ]
    query: "kas"
  ) {
    source {
      uri
      name
      creators {
        uri
        name
        alternateName
      }
    }
    result {
      __typename
      ... on Terms {
        terms {
          uri
          prefLabel
          altLabel
          hiddenLabel
          scopeNote
        }
      }
      ... on Error {
        message
      }
    }
  }
}
```

...then an error is returned: `Cannot read property 'id' of undefined`.

ddeboer:
Fix #143.

This is a result of #36.

### DIFF
--- a/src/services/terms.ts
+++ b/src/services/terms.ts
@@ -90,16 +90,26 @@ export class TermsTransformer {
         term.altLabels,
         term.hiddenLabels,
         term.scopeNotes,
-        term.broaderTerms.map(
-          iri => this.termsMap.get(iri.value) as RelatedTerm
-        ),
-        term.narrowerTerms.map(
-          iri => this.termsMap.get(iri.value) as RelatedTerm
-        ),
-        term.relatedTerms.map(
-          iri => this.termsMap.get(iri.value) as RelatedTerm
-        )
+        term.broaderTerms.reduce(this.mapRelatedTerms, []),
+        term.narrowerTerms.reduce(this.mapRelatedTerms, []),
+        term.relatedTerms.reduce(this.mapRelatedTerms, [])
       );
     });
   }
+
+  /**
+   * Map related IRIs to their related terms, making sure to only accept complete related terms.
+   *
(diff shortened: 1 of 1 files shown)


src/services/terms.ts
```diff
@@ -90,16 +90,26 @@ export class TermsTransformer {
...
     });
   }
+
+  /**
+   * Map related IRIs to their related terms, making sure to only accept complete related terms.
+   *
+   * Related terms can be incomplete because of the SPARQL query limit (see
+   * https://github.com/netwerk-digitaal-erfgoed/network-of-terms-api/issues/36).
+   */
+  private mapRelatedTerms = (acc: RelatedTerm[], iri: RDF.Term) => {
```
sdevalk, line 106: What does 'acc' mean?

sdevalk (commented):
Nice!

src/services/terms.ts
ddeboer: Good question! I’m used to call this argument `acc`, short for `accumulator` (from doing functional programming). 

As it turns out, [even JS calls it `accumulator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce). Would renaming this to `accumulator` help? Or is `terms`/`relatedTerms` clearer?

ddeboer: If you insist. 😁 Moved the reduce into the function.

sdevalk: 🕺

sdevalk (approved):
Let's go! :)

PR #210, opened 2021-07-05
feat: Add query timeout

### LINKED ISSUE #45: Abort request to slow or unreachable terminology source
A terminology source can be slow or unreachable. If it takes too much time to query the source, we should abort the request and inform the user about this. The default timeout of Comunica is 60 seconds, but this is too long for our use case. We should change this to e.g. 5 or 10 seconds.

ddeboer:
* Fix #45.
* Use `Promise.race` to set a timeout on each outgoing request.
* Configurable through `DEFAULT_QUERY_TIMEOUT` and `MAX_QUERY_TIMEOUT`
  env vars; respectively 5 and 10 sec.

### DIFF
--- a/src/commands/sources/query.ts
+++ b/src/commands/sources/query.ts
@@ -93,6 +93,7 @@ export class QuerySourcesCommand extends Command {
     const results = await service.queryAll({
       sources,
       query: flags.query,
+      timeoutMs: parseInt(process.env.QUERY_TIMEOUT as string) || 10000,
     });
     this.render(results, catalog);
   }
--- a/src/commands/sources/query.ts
+++ b/src/commands/sources/query.ts
@@ -34,6 +34,10 @@ export class QuerySourcesCommand extends Command {
       required: false,
       default: 'warn',
     }),
+    timeout: flags.integer({
+      description: 'Query timeout in ms',
+      required: false,
+    }),
   };
 
   protected render(results: TermsResult[], catalog: Catalog): void {
@@ -93,7 +97,7 @@ export class QuerySourcesCommand extends Command {
     const results = await service.queryAll({
       sources,
       query: flags.query,
-      timeoutMs: parseInt(process.env.QUERY_TIMEOUT as string) || 10000,
+      timeoutMs: flags.timeout,
     });
     this.render(results, catalog);
   }
--- a/src/server/resolvers.ts
+++ b/src/server/resolvers.ts
@@ -27,6 +27,7 @@ async function queryTerms(object: any, args: any, context: any): Promise<any> {
       (distributionIri: string) => new IRI(distributionIri)
     ),
     query: args.query,
+    timeoutMs: parseInt(process.env.QUERY_TIMEOUT as string) || 10000,
   });
   return results.map((result: TermsResult) => {
     if (result instanceof Error) {
--- a/src/server/resolvers.ts
+++ b/src/server/resolvers.ts
@@ -27,7 +27,7 @@ async function queryTerms(object: any, args: any, context: any): Promise<any> {
       (distributionIri: string) => new IRI(distributionIri)
     ),
     query: args.query,
-    timeoutMs: parseInt(process.env.QUERY_TIMEOUT as string) || 10000,
+    timeoutMs: args.timeoutMs,
   });
   return results.map((result: TermsResult) => {
     if (result instanceof Error) {
--- a/src/services/distributions.ts
+++ b/src/services/distributions.ts
@@ -1,6 +1,6 @@
 import * as Joi from '@hapi/joi';
 import Pino from 'pino';
-import {TermsResult, QueryTermsService} from './query';
+import {QueryTermsService, TermsResult} from './query';
 import {Catalog, IRI} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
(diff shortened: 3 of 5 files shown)


src/server/resolvers.ts
```diff
@@ -27,6 +27,7 @@ async function queryTerms(object: any, args: any, context: any): Promise<any> {
       (distributionIri: string) => new IRI(distributionIri)
     ),
     query: args.query,
+    timeoutMs: parseInt(process.env.QUERY_TIMEOUT as string) || 10000,
```
ddeboer, line 30: This means the timeout is configured on the server: clients have no way to provide their own timeout. I decided for this over the alternative where clients can provide their own desired timeout in queries for two reasons: 

1. I don’t want to burden clients with choosing some timeout.
2. We would still need some server-side upper bound so clients cannot have extremely long timeouts (e.g. 1000 sec).

However, this comes at the cost of some flexibility for clients. Should we assume they can all live with our default (10 sec) or do individual clients really need to set their own timeouts?

src/commands/sources/query.ts
```diff
@@ -93,6 +93,7 @@ export class QuerySourcesCommand extends Command {
     const results = await service.queryAll({
       sources,
       query: flags.query,
+      timeoutMs: parseInt(process.env.QUERY_TIMEOUT as string) || 10000,
```
sdevalk, line 96: Or we can retrieve the timeout value from the `flags` (similar to how we retrieve the `loglevel` value)? This allows a CLI user to set the timeout via the command line.

src/server/resolvers.ts
sdevalk: Yes, good point!

I'm not sure if all clients can live with our default. Perhaps some clients are willing to wait longer in case of slow(er) sources. That argues for a configuration option on the client. If that makes sense, we could even lower our default to, say, 3 to 5 seconds.

src/services/distributions.ts
```diff
@@ -19,21 +19,25 @@ const schemaConstructor = Joi.object({
 export interface QueryOptions {
   source: IRI;
   query: string;
+  timeoutMs: number;
 }
 
 const schemaQuery = Joi.object({
   source: Joi.object().required(),
   query: Joi.string().required(),
+  timeoutMs: Joi.number().required(),
```
sdevalk, line 28: We could add the timeout options to the Joi object, e.g.:

`timeoutMs: Joi.number().integer().min(1000).max(10000).default(10000)`

src/services/query.ts
```diff
@@ -116,3 +128,16 @@ const alphabeticallyByLabels = (a: Term, b: Term) => {
...
+
+function guardTimeout<T>(
+  promise: Promise<T>,
+  timeoutMs: number,
+  timeoutError: TimeoutError
+): Promise<T> {
+  return Promise.race([
+    promise,
+    new Promise(resolve =>
+      setTimeout(resolve.bind(null, timeoutError), timeoutMs)
```
sdevalk, line 140: Is it useful to log a (informational?) message in case of a timeout, to aid debugging?

sdevalk (changes requested):

src/services/query.ts
ddeboer: As the error message returned to clients is quite clear, we decided not to add logging for now:

```json
"result": {
  "__typename": "TimeoutError",
  "message": "Source timed out after 1ms"
}
```

src/commands/sources/query.ts
ddeboer: Yes, done!

src/server/resolvers.ts
ddeboer: Added a `timeoutMs` GraphQL query param.

src/services/query.ts
```diff
@@ -21,13 +21,18 @@ export interface ConstructorOptions {
...
+  timeoutMs: number;
 }
 
 const schemaConstructor = Joi.object({
   logger: Joi.object().required(),
   distribution: Joi.object().required(),
   query: Joi.string().required(),
   comunica: Joi.object().required(),
+  timeoutMs: Joi.number()
+    .integer()
```
sdevalk, line 33: Should we also define a minimum value with `min()` (e.g. 0 or 1000), to at least prevent nonsense values such as `-1`?

src/services/query.ts
```diff
@@ -21,13 +21,18 @@ export interface ConstructorOptions {
...
+    .max(parseInt(process.env.MAX_QUERY_TIMEOUT as string) || 10000)
```
sdevalk, line 34: Or `process.env.MAX_QUERY_TIMEOUT as number`?

sdevalk (changes requested):

src/services/query.ts
ddeboer: That doesn’t work:

```
TS2352: Conversion of type 'string | undefined' to type 'number' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.   Type 'string' is not comparable to type 'number'.
```

sdevalk: 👍🏻

sdevalk (approved):
Cool! Let's go! 🥳

PR #217, opened 2021-07-07
test: Add tests

### LINKED ISSUE #202: Add tests
The [catalog is tested separately](https://github.com/netwerk-digitaal-erfgoed/network-of-terms-catalog/tree/master/test). What kind of tests can we add for the API?

1. Start with high-level, end-to-end tests against the GraphQL endpoint to get as much coverage as quickly as possible (forgetting about error localisation for now).
2. Can we provide a mock catalog with a local endpoint so we don’t depend on the availability of sources for our tests to pass? 
3. The Register already has [some Fastify tests](https://github.com/netwerk-digitaal-erfgoed/register/blob/main/test/server.test.ts), so have a look at those.

ddeboer:
* Fix #202.
* Install Jest testing library and configure for TypeScript.
* Add code coverage requirements.
* Run on CI.


jest.config.js
```diff
@@ -0,0 +1,14 @@
...
+  roots: ['test/'],
+  preset: 'ts-jest',
+  collectCoverage: true,
+  coverageReporters: ['json-summary', 'text'],
+  coverageThreshold: {
+    global: {
+      lines: 94.65,
+      statements: 94.27,
+      branches: 89.29,
+      functions: 87.93,
```
ddeboer, line 11: Not too shabby. 😀

test/fixtures/terms.ttl
```diff
@@ -0,0 +1,22 @@
...
+    a skos:Concept ;
+    skos:prefLabel "Painting" ;
+    skos:related <https://example.com/resources/art> ;
+    skos:broader <https://example.com/resources/art> .
+
+<https://example.com/resources/art>
+    a skos:Concept ;
+    skos:prefLabel "Art" ;
+    skos:altLabel "All things art" ;
+    skos:narrower <https://example.com/resources/painting> .
```
ddeboer, line 22: Does this test data make enough sense so it won’t hinder maintenance in the future?

test/server/server.test.ts
```diff
@@ -0,0 +1,181 @@
...
+  IRI,
+  Organization,
+  SparqlDistribution,
+} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
+import {server} from '../../src/server/server';
+import {gql} from 'mercurius-codegen';
+import {setup, teardown} from 'jest-dev-server';
+
+let httpServer: FastifyInstance<Server>;
+describe('Server', () => {
```
ddeboer, line 15: When going through this file, do the test cases read well? Is it clear what is going on?

jest.config.js
```diff
@@ -0,0 +1,14 @@
...
+  roots: ['test/'],
+  preset: 'ts-jest',
+  collectCoverage: true,
+  coverageReporters: ['json-summary', 'text'],
+  coverageThreshold: {
+    global: {
+      lines: 95.31,
+      statements: 94.92,
+      branches: 98.33,
+      functions: 88.52,
```
ddeboer, line 11: Not too shabby. 😀

ddeboer:
The tests correctly captured the fact that GraphQL playground support was dropped, so please ignore that failure here (see #230 instead).

test/server/server.test.ts
sdevalk: I like it!

A small hesitation about the data structure on lines 16-70: it's not instantly clear to me what the meaning of each element is. For example:

```
    new IRI('https://example.com/datasets/timeout'),
    'Dataset the Second',
    [
      new Organization(
        new IRI('https://example.com/organizations/2'),
        'Organization the Second',
        'O2'
      ),
    ],
```

`Dataset the Second` is probably the name of the dataset and `O2` probably the alternate name of the organization. However, is there a way to make this clearer?

sdevalk (commented):
Nice work!

test/server/server.test.ts
ddeboer: I’m using the `Organization`, `Dataset` etc. classes from our `catalog` package. IntelliJ (and possibly other editors such as VSCode) displays the arguments, which improves readability:

Screenshot 2021-08-03 at 16 20 49

On the code level I see two possible improvements:

1. Use more realistic examples, e.g. `Koninklijke Bibliotheek`, `KB`.
2. Use named arguments, which in JS is only possible using object destructuring: `{name: 'Dataset the Second', alternateName: 'O2'}`.

ddeboer: Applied 1.

package.json
```diff
@@ -46,9 +54,10 @@
     "clean": "gts clean",
     "compile": "tsc",
     "fix": "gts fix",
-    "posttest": "npm run check",
-    "start": "node build/server/server.js",
-    "dev": "ts-node-dev src/server/server.ts",
+    "test": "jest ${1} --forceExit",
```
sdevalk, line 57: To be sure: do we need the `forceExit` flag? It could be an indication that the tear down in the tests isn't working properly.

test/server/server.test.ts
```diff
@@ -0,0 +1,219 @@
...
+  SparqlDistribution,
+} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
+import {server} from '../../src/server/server';
+import {gql} from 'mercurius-codegen';
+import {setup, teardown} from 'jest-dev-server';
+
+let httpServer: FastifyInstance<Server>;
+const catalog = new Catalog([
+  new Dataset(
+    new IRI('http://vocab.getty.edu/aat'),
```
sdevalk, line 17: Add support for multiple URI prefixes?

test/server/server.test.ts
```diff
@@ -0,0 +1,219 @@
...
+            creators {
+              uri
+              name
+              alternateName
+            }
+          }
+        }
+      `
+    );
+    expect(body.data['sources']).toHaveLength(catalog.datasets.length);
```
sdevalk, line 97: Out of curiosity: is `body.data.sources` allowed, rather than `body.data['sources']`?

sdevalk (changes requested):

package.json
ddeboer: Unfortunately, we do. Otherwise we run into:

```
Jest did not exit one second after the test run has completed.

This usually means that there are asynchronous operations that weren't stopped in your tests. Consider running Jest with `--detectOpenHandles` to troubleshoot this issue.
```

The only test case that throws this error is the one that tests our query timeout functionality. This is probably caused by https://github.com/comunica/comunica/issues/70.

test/server/server.test.ts
ddeboer: Actually, yes!

test/server/server.test.ts
ddeboer: This is the Dataset URI itself; not yet the URI prefix functionality, for which see #251.

sdevalk: Ah, I see! Perfect!

sdevalk (approved):
Excellent!

PR #230, opened 2021-07-13
Replace GraphQL playground

ddeboer:
The playground was removed in Mercurius 8.0
(https://github.com/mercurius-js/mercurius/pull/453). Instead, Mercurius
ships with GraphiQL, so enable that.

### DIFF
--- a/src/server/server.ts
+++ b/src/server/server.ts
@@ -35,6 +35,15 @@ async function startServer(): Promise<void> {
       reply.redirect('/graphiql');
     },
   });
+
+  // Redirect /playground to /graphiql for BC.
+  server.route({
+    method: 'GET',
+    url: '/playground',
+    handler: (req, reply) => {
+      reply.redirect(302, '/graphiql');
+    },
+  });
   await server.listen(3123, '0.0.0.0');
 }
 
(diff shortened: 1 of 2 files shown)


src/server/server.ts
```diff
@@ -18,7 +18,7 @@ async function startServer(): Promise<void> {
   server.register(mercurius, {
     schema,
     resolvers,
-    graphiql: 'playground',
+    graphiql: true,
```
ddeboer, line 21: Apparently a custom path is no longer possible.

sdevalk: Or should we use our own, 'persistent' path instead (e.g. stick with `/playground`)? If Mercurius decides to use another GraphQL editor again in the future, shipping with its own path, we too have to update our paths.

ddeboer: Apparently a custom path is no longer possible (`/playground` or `playground` didn’t work). But please note we’re still redirecting `/` to the playground. So as long as we refer developers to https://termennetwerk-api.netwerkdigitaalerfgoed.nl rather than https://termennetwerk-api.netwerkdigitaalerfgoed.nl/playground we’re fine.

sdevalk: Thanks! I was thinking about the use case where users have bookmarked `/playground` already. Will this change return a 404 when requesting this path? If so, would it be an option to not only redirect `/` to the playground, but also `/playground` (for backward compatibility)?

ddeboer: Added a redirect!

sdevalk (approved):
🕺

PR #251, opened 2021-07-26
feat: Add URI term lookup

### LINKED ISSUE #201: Look up terms by URI
Experimented a bit, and a query like the following, which combines both query search and URI lookup, seems to work:

```sparql
CONSTRUCT {
  ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2004/02/skos/core#Concept>.
  ?uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?prefLabel.
  ?uri <http://www.w3.org/2004/02/skos/core#altLabel> ?altLabel.
  ?uri <http://www.w3.org/2004/02/skos/core#hiddenLabel> ?hiddenLabel.
  ?uri <http://www.w3.org/2004/02/skos/core#scopeNote> ?scopeNote.
  ?uri <http://www.w3.org/2004/02/skos/core#broader> ?broader_uri.
  ?uri <http://www.w3.org/2004/02/skos/core#narrower> ?narrower_uri.
  ?uri <http://www.w3.org/2004/02/skos/core#related> ?related_uri.
  ?uri <http://www.w3.org/2004/02/skos/core#exactMatch> ?exactMatch.
  ?broader_uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?broader_prefLabel.
  ?narrower_uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?narrower_prefLabel.
  ?related_uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?related_prefLabel.
}
WHERE {
  {
    ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2004/02/skos/core#Concept>;
      <http://www.w3.org/2004/02/skos/core#inScheme> <http://data.bibliotheken.nl/id/scheme/brinkman>.
    OPTIONAL {
      ?uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?prefLabel.
      FILTER((LANG(?prefLabel)) = "nl")
    }
    OPTIONAL {
      ?uri <http://www.w3.org/2004/02/skos/core#altLabel> ?altLabel.
      FILTER((LANG(?altLabel)) = "nl")
    }
    OPTIONAL {
      ?uri <http://www.w3.org/2004/02/skos/core#hiddenLabel> ?hiddenLabel.
      FILTER((LANG(?hiddenLabel)) = "nl")
    }
    OPTIONAL {
      ?uri <http://www.w3.org/2004/02/skos/core#scopeNote> ?scopeNote.
      FILTER((LANG(?scopeNote)) = "nl")
    }
    OPTIONAL {
      ?uri <http://www.w3.org/2004/02/skos/core#broader> ?broader_uri.
      ?broader_uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?broader_prefLabel.
      FILTER((LANG(?broader_prefLabel)) = "nl")
    }
    OPTIONAL {
      ?uri <http://www.w3.org/2004/02/skos/core#narrower> ?narrower_uri.
      ?narrower_uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?narrower_prefLabel.
      FILTER((LANG(?narrower_prefLabel)) = "nl")
    }
    OPTIONAL {
      ?uri <http://www.w3.org/2004/02/skos/core#related> ?related_uri.
      ?related_uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?related_prefLabel.
      FILTER((LANG(?related_prefLabel)) = "nl")
    }
    OPTIONAL { ?uri <http://www.w3.org/2004/02/skos/core#exactMatch> ?exactMatch. }
    ?uri ?predicate ?label.
    VALUES ?predicate {
      <http://www.w3.org/2004/02/skos/core#prefLabel>
      <http://www.w3.org/2004/02/skos/core#altLabel>
    }
    FILTER(((LANG(?label)) = "nl") && (IF("" != "", bif:contains(?label, ""), false) || ?uri IN (<http://data.bibliotheken.nl/id/thes/p07568439X>)))
  }
}
LIMIT 1000
```

The important bit is:

```sparql
    FILTER(((LANG(?label)) = "nl") && (IF("" != "", bif:contains(?label, ""), false) || ?uri IN (<http://data.bibliotheken.nl/id/thes/p07568439X>)))
```

I also had to move the `VALUES` clause to the bottom of the query to prevent an – unexplained – `Virtuoso 37000 Error SP031: SPARQL compiler: No suitable triple pattern is found for a variable $label in special predicate bif:contains() at line 59 of query
`.

The question is: is this query still maintainable enough?

ddeboer:
Fix #201.

- [x] Depends on https://github.com/netwerk-digitaal-erfgoed/network-of-terms-catalog/pull/67.
- [x] Update readme.


src/server/schema.ts
```diff
@@ -33,10 +33,20 @@ export const schema = `
     terms: [Term]! @deprecated(reason: "Use 'result' instead")
     result: TermsResult!
   }
+  
+# TBD: Alternative result type.
+#  type LookupResult {
+#    uri: ID!
+#    source: Source
+#    result: LookupSuccessErrorResult
+#  }
```
ddeboer, line 42: TODO: discuss

src/services/query.ts
```diff
@@ -14,26 +14,19 @@ import Pino from 'pino';
...
 const schemaConstructor = Joi.object({
   logger: Joi.object().required(),
-  distribution: Joi.object().required(),
-  query: Joi.string().required(),
   comunica: Joi.object().required(),
-  timeoutMs: Joi.number()
-    .integer()
-    .min(1)
-    .max(parseInt(process.env.MAX_QUERY_TIMEOUT as string) || 10000)
-    .default(parseInt(process.env.DEFAULT_QUERY_TIMEOUT as string) || 5000),
```
ddeboer, line 36: Parameterised this arguments as part of the `lookup()` and `search()` functions instead.

src/services/query.ts
```diff
@@ -53,72 +46,92 @@ export class TimeoutError extends Error {
...
+      distribution,
+      timeoutMs,
+      Bindings({'?query': literal(searchQuery)})
+    );
+  }
+
+  async lookup(iris: IRI[], distribution: Distribution, timeoutMs: number) {
+    return this.run(
+      // TODO: insert real query here
+      'CONSTRUCT { ?s ?p ?o } WHERE { ?s ?p ?o . VALUES ?s { ?uris } }'.replace(
```
ddeboer, line 89: Didn’t find a way to use Comunica’s initialBindings in a `VALUES { ?initialBindings }` clause, so doing the replace myself.

sdevalk: 👍🏻

src/lookup/lookup.ts
```diff
@@ -0,0 +1,52 @@
...
+      },
+      new Map<Dataset, IRI[]>()
+    );
+
+    const lookups = <Promise<TermsResult>[]>[];
+    datasetToIris.forEach((iris, dataset) =>
+      lookups.push(this.query(dataset.distributions[0], iris, timeoutMs))
+    );
+
+    // TODO: decide whether we want to return results grouped by source or indexed by lookup URI.
```
sdevalk, line 33: Good one. My guess is that it's not very useful to group results by source. The API user is probably unaware of the sources, and wants to get to the metadata of each term as quickly as possible (without having to iterate over the sources first). An index by URI makes sense. But the source information would then become a part of the term information?

src/services/query.ts
sdevalk: 👍🏻

src/lookup/lookup.ts
ddeboer: Exactly: that would mean that we cannot re-use `TermsQueryResult` as the output of both the search and the lookup query:

```graphql
  type TermsQueryResult {
    source: Source!
    terms: [Term]! @deprecated(reason: "Use 'result' instead")
    result: TermsResult!
  }
```

Instead, we need a new output type for the lookup query, something like `LookupResult` below.

src/server/schema.ts
sdevalk: True. Or should we keep the `uri`, to allow API users to get to the right term faster? If they provide multiple URIs to the `lookup` endpoint, they need an easy way to find the metadata that belongs to each URI.

src/lookup/lookup.ts
sdevalk: Yes, `[LookupResult]` looks good!

src/server/schema.ts
ddeboer: Yes, I kept the URI. This is especially relevant when we can find no term for the URI, in which case `Term.uri` is not available.

src/lookup/lookup.ts
ddeboer: Done!

test/server/server.test.ts
```diff
@@ -134,6 +140,49 @@ describe('Server', () => {
     expect(body.data.terms[0].result.terms).toHaveLength(4); // Terms.
   });
 
+  it('responds to GraphQL lookup query', async () => {
```
ddeboer, line 143: This is a rather large PR, so you may want to start reading here, from the test case. Hopefully this shows the modalities relevant for query lookups.

src/server/schema.ts
```diff
@@ -22,37 +22,50 @@ export const schema = `
...
+
   type Terms {
     terms: [Term]
   }
-  
+
+  type LookupResult {
+    uri: ID!
+    source: Source
+    result: LookupSuccessErrorResult
```
ddeboer, line 52: `result` is null when we cannot find a term for the URI. Because this isn’t an error proper, I decided to make this nullable rather than add `TermNotFound` to the union type. Does that make sense? A downside is that the lookup can be (technically) successful (i.e., a dataset was found, its distribution could be reached and there were no query errors) but still return `null` (because no term was found).

This is related to the fact that I return a `LookupResult` for _all_ IRIs in the query, even those that returned `null`. I prefer to keep this over leaving out IRIs that didn’t return anything because it’s a nice guarantee for clients that _all_ the IRIs from their query will be contained in the result (even if no terms were found).

`source` is null when we cannot find a dataset whose prefix matches the term’s IRI. Alternatively, we could combine `result` and `source` together in a new type `LookupSuccessResult`.

README.md
```diff
@@ -192,3 +196,43 @@ query Terms {
   }
 }
 ```
+
+### Look up terms by URI
+
+Use the `lookup` query to look up terms whose URIs you know (for example, because you have stored the URIs previously):
+
+```graphql
+query {
```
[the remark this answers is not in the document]
ddeboer: Good point. Decided to remove the query name because it’s redundant.

src/server/schema.ts
ddeboer: Thought and read about this some more. It’s quite inelegant to have two nullable fields (both `source` and `result`). 

So how about this GraphQL schema and corresponding query structure instead?

Schema:

```graphql
type LookupQueryResult {
  uri: ID!
  source: SourceResult!
  result: LookupResult!
}

union SourceResult = Source | SourceNotFoundError

type SourceNotFoundError implements Error {
  message: String!
}

union LookupResult = Term | TimeoutError | ServerError | NotFoundError  

type TermNotFoundError implements Error {
  message: String!
}
```

Query:

```graphql
query {
  lookup(
    uris: ["https://data.rkd.nl/artists/32439", "https://data.cultureelerfgoed.nl/term/id/cht/15e29ea3-1b4b-4fb2-b970-a0c485330384"],
    timeoutMs: 1000
  ) {
    uri
    source {
      __typename
      ... on Source {
        source {
          uri
          name
          creators {
            uri
            name
            alternateName
          }
        }
      }
      ... on Error {
        message
      }
    }
    result {
      __typename
      ... on Term {
        term {
          uri
          prefLabel
          altLabel
          hiddenLabel
          scopeNote
          broader {
            uri
            prefLabel
          }
        }
      }
      ... on Error {
        message
      }
    }
  }
}`;
```

With this approach, we clearly separate the two steps in the lookup process: first find a source for the URI provided by the client, then query the source for that URI. My only gripe is that this might be too leaky an abstraction: are we bothering clients with details they aren’t interested in? Does the repeated inline fragment `... on Error { message }` overcomplicate the query?

ddeboer:
Implemented schema/query solution A, so ready for one final review.

src/lookup/lookup.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export type SourceResult = Distribution | SourceNotFoundError;
+
+export type LookupResult = Term | NotFoundError | TimeoutError | ServerError;
+
+export class SourceNotFoundError {
+  readonly message: string;
+
+  constructor(readonly iri: IRI) {
+    this.message = `No source found that can provide term with URI ${iri}`;
```
ddeboer, line 31: Is this message clear enough?

README.md
```diff
@@ -192,3 +196,48 @@ query Terms {
...
+        uri
+        name
+        creators {
+          uri
+          name
+          alternateName
+        }
+      }
+      ... on Error {
+        __typename
```
sdevalk, line 221: `__typename` is not present on line 238. Should we omit it here, or add it on line 238? Or should `__typename` be placed before `... on Source`?

src/lookup/lookup.ts
sdevalk: Perfect! :)

src/lookup/lookup.ts
```diff
@@ -0,0 +1,118 @@
...
+    const datasetToIris = [...irisToDataset].reduce(
+      (datasetMap, [iri, dataset]) => {
+        datasetMap.set(dataset, [...(datasetMap.get(dataset) ?? []), iri]);
+        return datasetMap;
+      },
+      new Map<Dataset, IRI[]>()
+    );
+
+    const lookups = <Promise<TermsResult>[]>[];
+    datasetToIris.forEach((iris, dataset) =>
```
sdevalk, line 70: Or can we use a `map` here? For example: `const lookups = datasetToIris.map(...)`

README.md
ddeboer: Good catch! Added the `__typename` to the `Error` object because it’s only relevant for narrowing down the error type; successful responses are just `Term`.

src/lookup/lookup.ts
ddeboer: Yes, that’s much cleaner. FYI, applying `.map` to a Map isn’t possible; you have to spread into an array first – which I overlooked and so used a `.forEach` where none was needed:

```javascript
[...datasetToIris].map(...)
```

PR #398, opened 2021-12-03
feat: Provide query variants to SPARQL query

### LINKED ISSUE #426: Searching on terms with diacritical marks
When I perform the query:

```
{"query":"query Terms ($sources: [ID]!, $query: String!) {\n                terms (sources: $sources query: $query) {\n                  source {\n                    name\n                    uri\n                    alternateName\n                    creators {\n                      name\n                      alternateName\n                    }\n                  }\n                  result {\n                    ... on Terms {\n                      terms {\n                        uri\n                        prefLabel\n                        altLabel\n                        hiddenLabel\n                        scopeNote\n                        seeAlso\n                        broader {\n                          uri\n                          prefLabel\n                        }\n                        narrower {\n                          uri\n                          prefLabel\n                        }\n                        related {\n                          uri\n                          prefLabel\n                        }\n                      }\n                    }\n                    ... on Error {\n                      __typename\n                      message\n                    }\n                  }\n                }\n              }","variables":{"sources":["https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql"],"query":"Joan Miró"}}
```

I get the error:

```
Invalid SPARQL endpoint response from https://api.data.netwerkdigitaalerfgoed.nl/datasets/rkd/rkdartists/services/rkdartists/sparql (HTTP status 500):Virtuoso 37000 Error XM029: Free-text expression, line 0: Invalid character in free-text search expression, it may not appear outside quoted string at �SPARQL query:define sql:big-data-const 0 define output:format "HTTP+TTL text/turtle" CONSTRUCT {  ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2004/02/skos/core#Concept>.  ?uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?schema_name.  ?uri <http://www.w3.org/2004/02/skos/core#altLabel> ?schema_alternateName.  ?uri <http://www.w3.org/2004/02/skos/core#scopeNote> ?schema_description.}WHERE {  {    ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?type.    VALUES ?type {      <http://schema.org/Person>      <http://schema.org/Organization>    }    ?uri ?name ?label.    VALUES ?name {      <http://schema.org/name>      <http://schema.org/alternateName>    }    OPTIONAL { ?uri <http://schema.org/name> ?schema_name. }    OPTIONAL { ?uri <http://schema.org/alternateName> ?schema_alternateName. }    OPTIONAL { ?uri <http://schema.org/description> ?schema_description. }    FILTER(<bif:contains>(?label, REPLACE(REPLACE("Joan Miró", "[.,]", " "), "(?<!AND)(?<!OR)[[:space:]]+(?!AND)(?!OR)(?!$)(?![[:space:]])", " AND ", "i")))  }}LIMIT 1000
```

It looks like terms with diacritical marks must be quoted when passed to `bif:contains`: manually testing with `url -H "Content-Type: application/sparql-query" -d '@/tmp/query.txt' https://api.data.netwerkdigitaalerfgoed.nl/datasets/rkd/rkdartists/services/rkdartists/sparql` and the following query:

```
CONSTRUCT
  {    ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> <http://www.w3.org/2004/02/skos/core#Concept>.
    ?uri <http://www.w3.org/2004/02/skos/core#prefLabel> ?schema_name.
    ?uri <http://www.w3.org/2004/02/skos/core#altLabel> ?schema_alternateName.
    ?uri <http://www.w3.org/2004/02/skos/core#scopeNote> ?schema_description.
  }
WHERE
  {
    {
      ?uri <http://www.w3.org/1999/02/22-rdf-syntax-ns#type> ?type.
      VALUES ?type {      <http://schema.org/Person>      <http://schema.org/Organization>    }
      ?uri ?name ?label.
      VALUES ?name {      <http://schema.org/name>      <http://schema.org/alternateName>    }
      OPTIONAL { ?uri <http://schema.org/name> ?schema_name. }
      OPTIONAL { ?uri <http://schema.org/alternateName> ?schema_alternateName. }
      OPTIONAL { ?uri <http://schema.org/description> ?schema_description. }
      FILTER(<bif:contains>(?label, "Joan AND 'Miró'"))
    }
  }
LIMIT 1000
```

does work, while without the quotes it produces the error similar to the one above.

ddeboer:
* Add GraphQL `queryMode` parameter: forwards-compatible
  `OPTIMIZED`, default `DEPRECATED` and optional `RAW`.
* Keep BC by defaulting to `DEPRECATED` during transition time. 
  After this, we can remove `DEPRECATED` and simplify our SPARQL queries.
* Extend current `?query` binding with a `?booleanQuery` binding geared towards Virtuoso query endpoints.
* Add documentation to README and GraphQL schema.
* Fix #426.


test/search/query-variants.test.ts
```diff
@@ -0,0 +1,48 @@
...
+        ['?booleanQuery', "'a' and 'b' AND 'c' or 'd'"],
+      ])
+    );
+  });
+
+  it('escapes quotation marks', () => {
+    expect(queryVariants("Rex Stewart's Big Eight")).toEqual(
+      new Map([
+        ['?query', "rex stewart's big eight"],
+        ['?booleanQuery', "'rex' AND 'stewart\\'s' AND 'big' AND 'eight'"],
```
ddeboer, line 44: Check: does SPARQL need double `\\` for escaping chars?

sdevalk: To be honest: I'm not sure about SPARQL. However, I do know that Virtuoso needs the double `\\` for escaping. See for example this NTA query: https://api.triplydb.com/s/oJd_r6Er5. The crucial part is `"'D\\'hondt'"`. If you omit one or two of the `\` before `'` then Virtuoso throws an error ("Unterminated single-quoted word or phrase").

sdevalk: Alternatively, we could use double quotes rather than single quotes. See for example this NTA query: https://api.triplydb.com/s/CuAAHslLZ, part `'"D\'hondt"'`.

sdevalk: Come to think of it: it could be that current (advanced) NoT users already provide the NoT with queries in which both quote types are being used, e.g. `"'D\\'hondt'"` or  `'"D\'hondt"'`. Should we add a check for this (similar to the check if boolean operators are already present), so as to prevent duplicate escaping (e.g. `'D\\'Hondt'` becomes `\\'D\\\\'Hondt\\'`)?

test/search/query-variants.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    expect(queryVariants('test')).toEqual(
+      new Map([
+        ['?query', 'test'],
+        ['?booleanQuery', "'test'"],
+      ])
+    );
+  });
+
+  it('transforms multiple word query', () => {
+    expect(queryVariants('Dr. H. Colijnstraat')).toEqual(
```
sdevalk, line 14: A rather nasty one is this search word: `Sammy Dowds & Leslie` (Muziekweb). This should result in `'Sammy' AND 'Dowds &' AND 'Leslie'`, not in `'Sammy' AND 'Dowds' AND '&' AND 'Leslie'` (Virtuoso will complain about the `'&'`: `phrase consists of noise words exclusively`). Alternatively, we could remove characters (or sequences of characters) that only consists of 'noise words' (e.g. `&` or `&&`).

sdevalk (commented):
Great work! I've added some remarks/considerations.

test/search/query-variants.test.ts
ddeboer: > To be honest: I'm not sure about SPARQL. However, I do know that Virtuoso needs the double `\\` for escaping. 

Good point: we’re talking about Virtuoso- rather than SPARQL-specifics here.

> Should we add a check for this (similar to the check if boolean operators are already present), so as to prevent duplicate escaping (e.g. `'D\\'Hondt'` becomes `\\'D\\\\'Hondt\\'`)?

Yes. I’ll change the code to:

1. prevent duplicate scanning by only escape not-yet-escaped single quotes (so `'` becomes `\\'` but `\\'` does not become `\\\\');
2. only quote not-yet-quoted words (so `hond` becomes `'hond'` but `'hond'` stays `'hond'` and `"hond"` stays `"hond"`).

test/search/query-variants.test.ts
ddeboer: The latter makes most sense. However, it seems [the set of noise words is configurable](http://docs.openlinksw.com/virtuoso/noisewords/), so how can we know which noise words to remove?

sdevalk: Ouch, I wasn't aware of the configurability of the noise words. So we cannot with confidence remove specific characters, such as `&`; a Virtuoso instance may have configured other noise words.

Could we continue with the first option, by not removing the noise word but by including it in the quoted part of a search word (e.g. `Dowds &`)? This is not a perfect solution, though: if a Virtuoso instance has configured `van` to be a noise word, and the search word of a NoT user is `van gogh`, our query pre-processor wil create the query `'van' AND 'gogh'`, resulting in the Virtuoso error `phrase consists of noise words exclusively`.

test/search/query-variants.test.ts
ddeboer: @sdevalk will check how double quotation marks `"` behave in Virtuoso. Do we have to escape them too?

ddeboer: Is this still relevant now that I’ve added a `RAW` query type here?

test/search/query-variants.test.ts
sdevalk: I think we still need a solution for this one: `Sammy Dowds & Leslie` - right?

`Dowds &` and `Dr.` are basically the same to Virtuoso: they both contain noise words, `&` and `.`, that are only allowed as part of a phrase that includes non-noise words. The difference, however, is that our `SMART` pre-processor shouldn't split the search words of a user on whitespace. Rather, it should keep non-noise (`Dowds`) and noise (` &`) together. Is this possible?

src/server/schema.ts
```diff
@@ -63,6 +66,20 @@ export const schema = `
...
+  enum QueryMode {
+    "This mode is only for backwards-compatibility. Switch to \`SMART\` mode for forwards-compatbility."
+    DEPRECATED @deprecated(reason: "This mode will be removed in the future. Use \`SMART\` instead.")
+    
+    "Connect search query words with a boolean AND operator and optimise query input for term sources."
+    SMART
+    
+    "Send the unaltered query input to the term sources. For advanced users that want to have full control over the value of the SPARQL parameter."
+    RAW
+  }
```
ddeboer, line 82: Are the above documentation strings easy to understand?

README.md
```diff
@@ -122,7 +122,11 @@ query Sources {
 ```graphql
 # Query Cultuurhistorische Thesaurus (CHT)
 query {
-  terms(sources: ["https://data.cultureelerfgoed.nl/PoolParty/sparql/term/id/cht"], query: "fiets") {
+  terms(
+    sources: ["https://data.cultureelerfgoed.nl/PoolParty/sparql/term/id/cht"],
+    query: "fiets",
+    queryMode: SMART   # Forwards-compatible query type.
```
ddeboer, line 128: 1. Is `queryMode` a good name for this parameter?
2. Is `SMART` a good name for its value, i.e. the forwards-compatible mode where pre-processing is done in TypeScript rather than SPARQL.

test/search/query-variants.test.ts
ddeboer: Changed this to just remove stop words. While this loses some precision in the query, which drops the ampersand completely and becomes `'sammy' AND 'dowds' AND 'Leslie'`, this makes the pre-processing easier to maintain for us.

README.md
sdevalk: 1. `queryMode` sounds good! Alternatively, `queryType`?
2. I'm okay with `SMART`. Alternatively, `AUTO` ('automatically'), a little less 'pompous'?

src/commands/sources/query.ts
```diff
@@ -37,10 +45,12 @@ export class QuerySourcesCommand extends Command {
     timeout: flags.integer({
       description: 'Query timeout in ms',
       required: false,
+      default: 10000,
     }),
   };
 
   protected render(results: TermsResult[], catalog: Catalog): void {
+    console.log('results', results);
```
sdevalk, line 53: Should this be here?

src/search/query-mode.ts
```diff
@@ -0,0 +1,57 @@
...
+      ]);
+    case QueryMode.DEPRECATED:
+      return new Map([['?query', query]]);
+  }
+}
+
+const stringQuery = (query: string) =>
+  query.toLowerCase().replace(/\s+/g, ' ').trim();
+
+const booleanQuery = (query: string) =>
```
sdevalk, line 27: A late thought: or should we rename `booleanQuery` to `virtuosoQuery`, to make very clear in which situation it can be used (and to prevent that the same variable is going to be used in a query for another triplestore, where it could yield unexpected results, e.g. because the `escape` and `filterStopWords` functions currently are Virtuoso-specific)?

src/server/schema.ts
```diff
@@ -63,6 +66,20 @@ export const schema = `
...
+  The mode in which the literal search query (\`query\`) is interpreted before it is sent to the term sources.
+  """
+  enum QueryMode {
+    "This mode is only for backwards-compatibility. Switch to \`SMART\` mode for forwards-compatbility."
+    DEPRECATED @deprecated(reason: "This mode will be removed in the future. Use \`SMART\` instead.")
+    
+    "Connect search query words with a boolean AND operator and optimise query input for term sources."
+    SMART
+    
+    "Send the unaltered query input to the term sources. For advanced users that want to have full control over the value of the SPARQL parameter."
```
sdevalk, line 80: Or perhaps not mention SPARQL? For example: in case of Virtuoso the `query` is not processed by the triplestore but send to Virtuoso's underlying full text search engine. Something like '...that want to have full control over the way a term source searches'?

src/server/schema.ts
```diff
@@ -63,6 +66,20 @@ export const schema = `
...
   }
+  
+  """
+  The mode in which the literal search query (\`query\`) is interpreted before it is sent to the term sources.
+  """
+  enum QueryMode {
+    "This mode is only for backwards-compatibility. Switch to \`SMART\` mode for forwards-compatbility."
+    DEPRECATED @deprecated(reason: "This mode will be removed in the future. Use \`SMART\` instead.")
+    
+    "Connect search query words with a boolean AND operator and optimise query input for term sources."
```
sdevalk, line 77: Or we could generalize this a bit by merely saying that this query mode takes care of query optimization, without mentioning its details (e.g. that it uses `AND` - that could change if we, at some point, decide to query Virtuoso differently)?

sdevalk (changes requested):
Nice!

src/commands/sources/query.ts
ddeboer: Definitely not; removed!

src/server/schema.ts
ddeboer: Good point; removed the reference to SPARQL.

src/search/query-mode.ts
ddeboer: Agreed; changed to `virtuosoQuery`. I originally had `booleanQuery` to sort of hide the fact that we’re optimising for a specific triple store, but yeah, after all that’s really what we’re doing.

README.md
ddeboer: 1. I previously had `queryType` but considered the word `type` confusing in the context of a GraphQL schema because in GraphQL (query) types mean something else.
2. Yeah, `SMART` sounds smarter than it really is. 😁 `AUTO`, on the other hand, is somewhat vague. Perhaps `OPTIMIZED`?

ddeboer: Renamed to `OPTIMIZED`.

PR #468, opened 2022-02-07
fix: remove Pino types

sdevalk:
Pino nowadays provides its own types out of the box


ddeboer:
Thanks! Added the lock file, so this should be good to merge now.

Issue #265, opened 2021-08-12
Reconciliation API


ddeboer:
https://docs.google.com/document/d/1AUeMAiBW_gNiWZM-OYJW-iY_zaDbalEriIPiDBxuQW4

https://reconciliation-api.github.io/specs/latest/

https://docs.openrefine.org/manual/reconciling

### Questions

- [x] Should we call each source simply by its name (e.g. ‘RKDartists’) or make it clear that the reconciliation is provided not by the source itself but by us (e.g. `RKDartists via Termennetwerk`)?
- [ ] Should we add our reconciliation sources to the [list of public endpoints](https://reconciliation-api.github.io/testbench/)? 
- [x] Should we have another way for clients to discover our reconciliation service URLs? **List of URLs is now available at https://termennetwerk.netwerkdigitaalerfgoed.nl/faq.**

### Progress

- [x] #476 
- [x] #479 
- [ ] #266

PR #476, opened 2022-02-07
feat: Add Reconciliation service

ddeboer:
* Add feature to current API repository because that’s the fastest way
  to get started. Later, we may want to restructure the repositories.
* Reorganize tests to filter on query text for better assertions.

#265.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -3,7 +3,10 @@ import {server} from './server/server';
 
 (async () => {
   try {
-    const httpServer = await server(await Catalog.default());
+    const httpServer = await server(await Catalog.default(), [
+      // Only a single reconciliation service for now.
+      'https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql',
+    ]);
     await httpServer.listen(3123, '0.0.0.0');
   } catch (err) {
     console.error(err);
--- a/src/main.ts
+++ b/src/main.ts
@@ -4,8 +4,9 @@ import {server} from './server/server';
 (async () => {
   try {
     const httpServer = await server(await Catalog.default(), [
-      // Only a single reconciliation service for now.
+      // Only a couple of reconciliation services to begin with.
       'https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql',
+      'https://data.cultureelerfgoed.nl/PoolParty/sparql/term/id/cht',
     ]);
     await httpServer.listen(3123, '0.0.0.0');
   } catch (err) {
--- a/src/reconciliation/manifest.ts
+++ b/src/reconciliation/manifest.ts
@@ -1,0 +1,27 @@
+import {Catalog, IRI} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
+
+export class ServiceManifest {
+  public readonly versions = ['0.1', '0.2'];
+  public readonly schemaSpace = 'http://www.w3.org/2004/02/skos/core#Concept';
+
+  constructor(
+    public readonly name: string,
+    public readonly identifierSpace: IRI
+  ) {}
+}
+
+export function findManifest(
+  distributionIri: IRI,
+  catalog: Catalog
+): ServiceManifest | undefined {
+  const source = catalog.getDatasetByDistributionIri(distributionIri);
+  if (source && reconciliationIris.includes(distributionIri.toString())) {
+    return new ServiceManifest(source.name, source.iri);
+  }
+
+  return undefined;
+}
+
+const reconciliationIris = [
+  'https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql', // Start with a single reconciliation source.
+];
--- a/src/reconciliation/manifest.ts
+++ b/src/reconciliation/manifest.ts
@@ -12,16 +12,13 @@ export class ServiceManifest {
 
 export function findManifest(
   distributionIri: IRI,
-  catalog: Catalog
+  catalog: Catalog,
+  reconciliationServices: string[]
 ): ServiceManifest | undefined {
   const source = catalog.getDatasetByDistributionIri(distributionIri);
-  if (source && reconciliationIris.includes(distributionIri.toString())) {
+  if (source && reconciliationServices.includes(distributionIri.toString())) {
     return new ServiceManifest(source.name, source.iri);
   }
 
   return undefined;
 }
-
-const reconciliationIris = [
-  'https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql', // Start with a single reconciliation source.
-];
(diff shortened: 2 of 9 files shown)


src/reconciliation/manifest.ts
```diff
@@ -0,0 +1,24 @@
+import {Catalog, IRI} from '@netwerk-digitaal-erfgoed/network-of-terms-catalog';
+
+export class ServiceManifest {
+  public readonly versions = ['0.1', '0.2'];
+  public readonly schemaSpace = 'http://www.w3.org/2004/02/skos/core#Concept';
```
ddeboer, line 5: Is this correct?

src/reconciliation/query.ts
```diff
@@ -0,0 +1,58 @@
...
+        queryString,
+        QueryMode.OPTIMIZED,
+        distribution,
+        10000
+      );
+      const terms = termsResult instanceof Terms ? termsResult.terms : [];
+      results[queryId] = {
+        result: terms.map(term => ({
+          id: term.id.value.toString(),
+          name: term.prefLabels.map(label => label.value).join(' • '), // Join similarly to network-of-terms-demo.
```
ddeboer, line 36: Does this make sense? How do we want prefLabels to be shown in the reconciliation service?

test/server/server.test.ts
```diff
@@ -218,6 +236,52 @@ describe('Server', () => {
       expect(response.headers.location).toEqual('/graphiql');
     }
   );
+
+  it('returns reconciliation service manifest', async () => {
+    const response = await httpServer.inject({
+      method: 'GET',
+      url: '/reconciliation/https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql',
```
ddeboer, line 243: I put the whole distribution IRI after `/reconciliation/` because it seems the service URL is not allowed to have parameters:

> When the reconciliation service [endpoint](https://reconciliation-api.github.io/specs/latest/#dfn-endpoint) is queried with a HTTP GET query without parameters, the [service manifest](https://reconciliation-api.github.io/specs/latest/#dfn-service-manifest) MUST be returned.

Am I right in interpreting this as _query_ parameters? Otherwise, `/reconciliation?iri=https://...` would be a possibility too.

src/main.ts
```diff
@@ -3,7 +3,10 @@ import {server} from './server/server';
 
 (async () => {
   try {
-    const httpServer = await server(await Catalog.default());
+    const httpServer = await server(await Catalog.default(), [
+      // Only a single reconciliation service for now.
+      'https://data.netwerkdigitaalerfgoed.nl/rkd/rkdartists/sparql',
```
ddeboer, line 8: Inject enabled reconciliation services into the server (so this could come from an env var or secondary catalog later) but hard-code it for now.

src/reconciliation/manifest.ts
sdevalk: Interesting :) The spec appears to only allow one URI. But what if a recon service uses more than one? The NoT mostly uses SKOS, but we also have a bit of RDFS (and perhaps we need another ontology in the future). We could point to a general page with info about our schema? ('This URI also MAY resolve to a page describing these properties and their identifiers.')

src/reconciliation/query.ts
sdevalk: Good one. Our data model allows for more than one prefLabel (which could be handy if we want to start supporting multiple languages). But I'm not sure if there is a source that actually has 2+ prefLabels for a given term (in the same language). Anyway, does it make sense to return all the prefLabels - if any - in the `name`? Yes, I'd say so: if a source has multiple prefLabels, there's probably a reason for this. The `•` as a separator is interesting. I would prefer something more standardized (CSV?) so that the value of `name` could be parsed by a client, if desired. Perhaps `; `? But we'll run into problems if a prefLabel contains this separator :)

src/reconciliation/query.ts
```diff
@@ -0,0 +1,58 @@
...
+          score: 1, // Hard-coded score for now because scoring will have to be done client-side.
```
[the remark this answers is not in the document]
sdevalk: To be sure: I'd think that a default doesn't suffice and that our recon service has to calculate a score, not the client?

src/server/server.ts
```diff
@@ -4,18 +4,24 @@ import * as Logger from '../helpers/logger';
...
-  catalog: Catalog
+  catalog: Catalog,
+  reconciliationServices: string[]
 ): Promise<FastifyInstance<Server>> {
   const logger = Logger.getHttpLogger({
     name: 'http',
     level: 'info',
   });
   const comunica = await newEngine();
+  const queryTermsService = new QueryTermsService({comunica, logger: logger});
```
sdevalk, line 24: Minor: you could probably just say `logger` (instead of `logger: logger`)?

src/server/server.ts
```diff
@@ -47,5 +54,47 @@ export async function server(
     },
   });
 
+  server.get<{Params: {'*': string}}>('/reconciliation/*', (request, reply) => {
```
sdevalk, line 57: Minor: The Getty uses `/reconcile` and the unofficial VIAF and GeoNames services too. I'm not sure what's best practice here (it's just a name, but we could go for the most common one).

src/server/server.ts
```diff
@@ -47,5 +54,47 @@ export async function server(
...
+        distributionIri,
+        catalog,
+        reconciliationServices
+      );
+      if (manifest === undefined) {
+        reply.code(404).send();
+        return;
+      }
+
+      // Reconciliation queries are JSON-encoded in a x-www-form-urlencoded ‘queries’ parameter.
```
sdevalk, line 85: Ouch! :) I'm not sure why the spec doesn't allow for `application/json`

src/reconciliation/manifest.ts
ddeboer: Based on the specs, we decided to keep this for now:

> A URI which identifies the type of an [entity](https://reconciliation-api.github.io/specs/latest/#dfn-entity) returned by the reconciliation service, for example http://www.w3.org/2004/02/skos/core#Concept or https://schema.org/Thing. This URI MAY resolve to a page describing this type.

test/server/server.test.ts
ddeboer: Make this `/reconcile/.…`

src/main.ts
ddeboer: Add another source: CHT.

src/reconciliation/query.ts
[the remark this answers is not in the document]
ddeboer: Improved explanation.

src/server/server.ts
ddeboer: Changed URL to `/reconcile`.

sdevalk (approved):
Let's go! :)

PR #479, opened 2022-02-10
feat: Add HTML previews

ddeboer:
* Translate labels in preview based on browser’s Accept-Language header.

#265.﻿


src/reconciliation/preview.ts
```diff
@@ -0,0 +1,50 @@
...
+      <p>${literal(term.scopeNotes)}</p>
+      <dl>
+        ${
+          term.altLabels.length > 0
+            ? `<dt>Alternatieve labels</dt><dd>${literal(term.altLabels)}</dd>`
+            : ''
+        }
+        ${relatedTerms('Bredere termen', term.broaderTerms)}
+        ${relatedTerms('Nauwere termen', term.narrowerTerms)}
+        ${relatedTerms('Gerelateerde termen', term.relatedTerms)}
```
ddeboer, line 29: Haven’t found a way in the Reconciliation spec to parameterise the language, so decided to use Dutch here. Would English make more sense?

src/server/server.ts
```diff
@@ -37,6 +41,13 @@ export async function server(
   });
   server.register(fastifyCors);
   server.register(formBodyPlugin);
+  server.decorateRequest('previewUrl', '');
+  server.addHook('onRequest', (request, reply, done) => {
+    request.raw.previewUrl =
+      request.protocol + '://' + request.hostname + '/preview/{{id}}';
+    done();
+  });
```
ddeboer, line 49: It seems the preview URL must be absolute, so we have to dynamically get the request’s hostname to build the URL.

src/reconciliation/preview.ts
ddeboer: Good suggestion! 👍

OpenRefine indeed sends an `Accept-Language` header with preview requests. By inspecting the headers as they arrived at our preview service, I found that `Accept-Language`’s value depends not on the OpenRefine language setting but on the browser’s locale. While not ideal (Dutch users may have their browser set to English), we can respond to the value in our preview service.

sdevalk: Excellent!

Issue #492, opened 2022-02-18
GTAA lookup doesn’t work


ddeboer:
For example https://termennetwerk.netwerkdigitaalerfgoed.nl/lookup?uri=http://data.beeldengeluid.nl/gtaa/34685

ddeboer:
This is caused by the fact that the three GTAA datasets are in separate graphs, each with their own SPARQL endpoint, but with the same URI prefix. In other words: the lookup happens to use the first in line (genres) for all lookups. Because of this, only genres but not the other two types (geographical names and subjects will be found).

ddeboer:
Partly fixed in #515, but we can’t yet guarantee that the correct source is returned in the lookup result.

One way to solve this would be to map an extra property (e.g. `skos:inScheme`) that lookup queries can fill with the dataset URI, the inverse of #531.

Issue #491, opened 2022-02-17
Use void:uriSpace instead of schema:url


ddeboer:
For the term URI prefix we currently use `schema:url`, but `void:uriSpace` is a better fit. Is this precision worth it to add an extra vocabulary to our dataset definitions?

PR #515, opened 2022-02-24
Use new GTAA SPARQL endpoint

ddeboer:
- added one GTAA dataset for testing
- added query variable as filter


packages/network-of-terms-catalog/catalog/datasets/gtaa_persoonsnamen.jsonld
```diff
@@ -0,0 +1,37 @@
...
+      "@value": "GTAA Persoonsnamen"
+    }
+  ],
+  "creator": [
+    {
+      "@id": "https://www.beeldengeluid.nl/"
+    }
+  ],
+  "url": [
+    "http://data.beeldengeluid.nl/gtaa/Persoonsnamen"
```
ddeboer, line 17: Is this indeed correct, that is do all Persoonsnamen concepts begin with this URL prefix? If so, this will solve #492.

packages/network-of-terms-catalog/catalog/datasets/genre_filmmuseum.jsonld
```diff
@@ -0,0 +1,37 @@
+{
```
ddeboer, line 1: Please use dashes `-` the filenames, as we already had for the old version of the GTAA endpoint: `genre-filmmuseum.jsonld`.

ddeboer (commented):
Please make sure the filenames of the `.rq` files match those in the `jsonld` data. Use dashes `-` (not underscore `_`) everywhere for consistency.

packages/network-of-terms-catalog/catalog/datasets/gtaa_onderwerpen_beng.jsonld
```diff
@@ -0,0 +1,37 @@
+{
+  "@context": "https://schema.org/docs/jsonldcontext.jsonld",
+  "@id": "http://data.beeldengeluid.nl/gtaa/OnderwerpenBenG",
+  "@type": "Dataset",
+  "name": [
+    {
+      "@language": "nl",
+      "@value": "GTAA Onderwerpen beeld-geluid"
```
ddeboer, line 8: This is a slightly weird name.

packages/network-of-terms-catalog/catalog/datasets/gtaa_namen.jsonld
```diff
@@ -0,0 +1,37 @@
+{
+  "@context": "https://schema.org/docs/jsonldcontext.jsonld",
+  "@id": "http://data.beeldengeluid.nl/gtaa/Namen",
+  "@type": "Dataset",
+  "name": [
+    {
+      "@language": "nl",
+      "@value": "GTAA Namen"
```
ddeboer, line 8: We tend to use colon `:` to indicate subsets, so we had `GTAA: geografische namen` for example. Does that make sense or is the official name really `GTAA Namen`?

packages/network-of-terms-catalog/catalog/datasets/genre_filmmuseum.jsonld
```diff
@@ -0,0 +1,37 @@
+{
+  "@context": "https://schema.org/docs/jsonldcontext.jsonld",
+  "@id": "http://data.beeldengeluid.nl/gtaa/Genre_Filmmuseum",
+  "@type": "Dataset",
+  "name": [
+    {
+      "@language": "nl",
+      "@value": "Genre Filmmuseum"
```
ddeboer, line 8: Should `GTAA` be part of this name? `Genre Filmmuseum` looks really weird. Did you mean: `Genres Filmmuseum`?

ddeboer:
I cannot test these queries locally (because I have no access to the SPARQL endpoint), so I’m assuming they are correct.

packages/network-of-terms-catalog/catalog/datasets/gtaa_persoonsnamen.jsonld
ddeboer: This does not seem to be the case, e.g. https://data.beeldengeluid.nl/gtaa/30369 of Genre Filmmuseum does not start with `http://data.beeldengeluid.nl/gtaa/Genre_Filmmuseum` (the URL that is defined below for this dataset).

wmelder: Concept URI's all start with the same prefix: http://data.beeldengeluid.nl/gtaa/{some_id}
So, Person names don't have a different prefix that geographical names, for example.

packages/network-of-terms-catalog/catalog/datasets/genre_filmmuseum.jsonld
wmelder: Is it ok that I remove the older GTAA files and replace them with the newer ones, including dashes?

packages/network-of-terms-catalog/catalog/datasets/gtaa_onderwerpen_beng.jsonld
wmelder: That is what it is referred to. And indeed confusing. It has nothing to do with the Instituut voor Beeld en Geluid, but its just that terms from these lists are more focussing on things that can be seen (beeld) or can be heard (sound). The ohter Onderwerpen axis is abstracting away from physical and concrete stuff. E.g. When you see a 'Suzuki' the more general subject will be 'Cars'. The beeld-geluid axis would sooner accept more specific instances of concepts. I hope this example helps.

packages/network-of-terms-catalog/catalog/datasets/gtaa_namen.jsonld
wmelder: I will update conform your suggested format.

packages/network-of-terms-catalog/catalog/datasets/genre_filmmuseum.jsonld
ddeboer: Yes!

packages/network-of-terms-catalog/catalog/datasets/gtaa_persoonsnamen.jsonld
ddeboer: The dataset’s `schema:url` must be its URL prefix, so that would be `http://data.beeldengeluid.nl/gtaa/` for all datasets, if I understand you correctly. This prefix is used for looking up terms by their URI. Unfortunately, if all datasets share the same prefix, #492 will not be solved. Is there perhaps a generic SPARQL endpoint for lookups in all datasets, or do lookups need to go to the individual endpoints?

(Our property name `schema:url` is slightly unclear, but we decided to stick with pure Schema.org for now, even if it’s less expressive. See also #491.)

packages/network-of-terms-catalog/catalog/datasets/filmmuseum-genres.rq
```diff
@@ -0,0 +1,37 @@
...
+  "distribution": [
+    {
+      "@id": "https://data.beeldengeluid.nl/id/datadownload/0033",
+      "@type": "DataDownload",
+      "contentUrl": "https://gtaa.apis.beeldengeluid.nl/sparql",
+      "encodingFormat": "application/sparql-query",
+      "potentialAction": [
+        {
+          "@type": "SearchAction",
+          "query": "file://catalog/queries/search/genre-filmmuseum.rq"
```
ddeboer, line 28: The actual file names still have underscores, causing the automated tests to fail on GH Actions.

packages/network-of-terms-catalog/catalog/datasets/genre_filmmuseum.jsonld
ddeboer: It’s okay to add non-GTAA datasets, of course. I’m just wondering if this dataset is _only_ being used by Filmmuseum or are other institution also using it or going to use it?

packages/network-of-terms-catalog/catalog/datasets/filmmuseum-genres.rq
```diff
@@ -0,0 +1,37 @@
...
+    {
+      "@id": "https://www.beeldengeluid.nl/"
+    }
+  ],
+  "url": [
+    "http://data.beeldengeluid.nl/gtaa/Genre_Filmmuseum"
+  ],
+  "distribution": [
+    {
+      "@id": "https://data.beeldengeluid.nl/id/datadownload/0033",
```
ddeboer, line 21: Should these IRIs be resolvable? Currently they are not.

packages/network-of-terms-catalog/catalog/datasets/gtaa_onderwerpen_beng.jsonld
ddeboer: It does. The name still needs the colon `:` for consistency.

packages/network-of-terms-catalog/catalog/datasets/filmmuseum-genres.rq
wmelder: yes, now they are. the data catalog needed to be updated.

packages/network-of-terms-catalog/catalog/datasets/filmmuseum-genres.rq
wmelder: they are not in the local branch. I just removed the dataset files that had underscores and pushed. query files with underscores I don't see.

packages/network-of-terms-catalog/catalog/datasets/genre_filmmuseum.jsonld
wmelder: I only know of EYE as a user of Genres Filmmuseum.

packages/network-of-terms-catalog/catalog/datasets/gtaa_onderwerpen_beng.jsonld
wmelder: It has a colon locally. Can you check again? Maybe it was fixed in the latest push. We should be in sync.

packages/network-of-terms-catalog/catalog/datasets/gtaa_persoonsnamen.jsonld
wmelder: It doesn't seem right to use the schema:url property for the prefix. The schema:url is the address for the resource. Because the GTAA parts are organised around ConceptSchemes, In our case the ConceptScheme address is the resource. 

But, one could also argue that the Dataset is the resource, and therefore the schema:url should be
``  "url": [
    "https://data.beeldengeluid.nl/id/dataset/0010"
  ],
`` and maybe the ID should also be:
``
  "@id": "https://data.beeldengeluid.nl/id/dataset/0010",
  "@type": "Dataset",
``

I would prefer to use the dataset ID as the schema:``@id``, and the ConceptScheme URI as schema:url

``
  "@id": "https://data.beeldengeluid.nl/id/dataset/0010",
  "@type": "Dataset",
``
``  "url": [
    "http://data.beeldengeluid.nl/gtaa/OnderwerpenBenG"
  ],
`` 

But, then there is no prefix for lookup. Why is that necessary anyway? The Concepts all have a proper URI, so why would you need a prefix? For your information: the GTAA Concepts are all dereferanceable using the URI. Isn't that enough for lookup? Also, no basic auth is required for that.

packages/network-of-terms-catalog/catalog/datasets/filmmuseum-genres.rq
```diff
@@ -0,0 +1,37 @@
+{
```
ddeboer, line 1: This file should have extension jsonld, not rq.

wmelder: sorry. that was a bit sloppy of me. should be fixed.

packages/network-of-terms-catalog/catalog/datasets/gtaa_persoonsnamen.jsonld
wmelder: Sorry, my response might have unintentionally sounded a bit agitated at the end. I just wondered why you need the prefix.

ddeboer: I’ve just [added some documentation](https://github.com/netwerk-digitaal-erfgoed/network-of-terms/tree/master/packages/network-of-terms-catalog#data-model) to explain the need for the prefix. The Network of Terms has its own lookup functionality to present the same SKOS representation to clients as when they search for terms. Lookup is needed because:

- some sources dereference URIs correctly but return it in their own vocabulary, such as Schema.org, rather than SKOS, which clients expect from the Network of Terms
- other sources use SKOS internally but fail to dereference URIs, including GTAA (as of now):

      $ curl -v -H Accept:application/ld+json https://data.beeldengeluid.nl/gtaa/30369
      < HTTP/1.1 500 Internal Server Error
      
      Could not find acceptable representation

- even if sources dereference correctly and use SKOS, they may return values different from those returned by the Network of Terms search, which can be confusing for clients.

To know which term to source for a certain URI, the Network of Terms needs to know those prefixes.

Does that clear things up?

wmelder:
The GTAA URIs have HTTP as protocol, not https. They can be dereferenced without any authentication.



Verstuurd vanaf mijn iPhone

wmelder:
The Genre_Filmmuseum dataset can be left out of the network of terms for now. I will prepare a commit for that.

packages/network-of-terms-catalog/catalog/datasets/gtaa_persoonsnamen.jsonld
wmelder: Please note that the GTAA URI's all have the HTTP protocol. We should provide a better return code when using HTTPS for dereferencing.

``To know which term to source for a certain URI, the Network of Terms needs to know those prefixes.``
I still don't get it. Why do you need the prefix? For all sources it is different what the path after that prefix is. In our case the skos:notation value is always corresponding to it: ``/gtaa/{skos:notation}``. But that is not what is needed to dereference the concept. You can use the whole Concept URI (always HTTP) to get the data. Also, there is an API that can be used to get the concept data, but that is not involved at all in the network of terms. Given the queries in ``catalog/queries/lookup/`` and the documentation that you added: ``lookup queries retrieve a single term based on its URI.`` I would deduct that the SPARQL endpoint is used to get the Concept data. Why define a prefix then? It is just our GTAA namespace, not an API function.

ddeboer: > Please note that the GTAA URI's all have the HTTP protocol.

Ah, good to know! However, `curl -v -L -H Accept:application/ld+json http://data.beeldengeluid.nl/gtaa/30369` still returns a 500 error.

Anyway, the lookup feature works like this. On https://termennetwerk.netwerkdigitaalerfgoed.nl/lookup, a user enters any term’s URI, for example `http://vocab.getty.edu/aat/300046300` or `http://data.beeldengeluid.nl/gtaa/30369`. 

The question is: how does the Network of Terms know _which SPARQL endpoint to consult_ for each URI? This is what we need the URI prefix for:

- `http://vocab.getty.edu/aat/...` goes to the AAT endpoint;
- `http://data.beeldengeluid.nl/gtaa/...` to the GTAA one.

Does that make sense?

wmelder: >Ah, good to know! However, curl -v -L -H Accept:application/ld+json http://data.beeldengeluid.nl/gtaa/30369 still returns a 500 error.

Sorry to hear that. Now that you bring it up, I realize that I brought up that issue some time ago. It is acknowledged as a bug that it doesn't provide JSON-LD. It will give you the XML representation: ``curl  -L -H Accept:application/rdf+xml http://data.beeldengeluid.nl/gtaa/30369

wmelder: >if the URI begins with http://data.beeldengeluid.nl/gtaa/..., retrieve it from the GTAA endpoint (and so on).
Does that make sense?

I was under the impression that the lookup function was a type of query 'per source'. Now I understand that you do some delegation of the concept URI dereferencing 'by using the schema prefix'. In that case, for any of the GTAA sources it holds that any GTAA URI can be resolved and then ``http://data.beeldengeluid.nl/gtaa`` is the prefix for ALL GTAA sources. And they all use the same SPARQL endpoint, so all concepts are in all endpoints. 
The difficulty for the network of terms is that if many sources share the same prefix, which one does it need to choose?
Luckily, the solution is easy: If we make sure all GTAA lookup queries are not filtering on ConceptScheme it doesn't matter what source you choose: any GTAA source is good.
So, if you agree that this is the solution for the network of terms, than I will make the changes and commit. This means I remove all the triples like this: ``    ?uri skos:inScheme <http://data.beeldengeluid.nl/gtaa/GeografischeNamen> . ``

ddeboer: Please go ahead! You can then also reduce the lookup queries to a single one, and use that from all datasets. (I’ve already done this for the search queries.)

wmelder: All GTAA lookup queries share the same file now.

wmelder:
After merging master into branch these errors arise when running `npm dev`:
>10:55:51 AM - File change detected. Starting incremental compilation...
../network-of-terms-catalog/src/catalog.ts(11,9): error TS2595: 'globby' can only be imported by using a default import.
../network-of-terms-catalog/src/catalog.ts(104,15): error TS7006: Parameter 'file' implicitly has an 'any' type.
10:55:52 AM - Found 2 errors. Watching for file changes.

ddeboer:
And… merged. Thanks a lot for your contribution and patience @wmelder!

PR #536, opened 2022-03-03
feat: Enable Reconciliation API for all relevant sources

ddeboer:
* Enable Reconciliation API in catalog instead of hardcoding. 

Ref #265﻿

### DIFF
--- a/packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
+++ b/packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
@@ -36,6 +36,13 @@
         {
           "@type": "FindAction",
           "query": "file://catalog/queries/lookup/abr-cht.rq"
+        },
+        {
+          "@type": "Action",
+          "actionApplication": {
+            "@id": "https://reconciliation-api.github.io/specs/latest/",
+            "@type": "SoftwareApplication"
+          }
         }
       ]
     }
--- a/packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
+++ b/packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
@@ -39,9 +39,12 @@
         },
         {
           "@type": "Action",
-          "actionApplication": {
-            "@id": "https://reconciliation-api.github.io/specs/latest/",
-            "@type": "SoftwareApplication"
+          "target": {
+            "@type": "EntryPoint",
+            "actionApplication": {
+              "@id": "https://reconciliation-api.github.io/specs/latest/",
+              "@type": "SoftwareApplication"
+            }
           }
         }
       ]
--- a/packages/network-of-terms-catalog/catalog/datasets/gtaa-classificatie.jsonld
+++ b/packages/network-of-terms-catalog/catalog/datasets/gtaa-classificatie.jsonld
@@ -30,6 +30,13 @@
         {
           "@type": "FindAction",
           "query": "file://catalog/queries/lookup/gtaa.rq"
+        },
+        {
+          "@type": "Action",
+          "actionApplication": {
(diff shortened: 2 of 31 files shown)


packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
```diff
@@ -36,6 +36,13 @@
         {
           "@type": "FindAction",
           "query": "file://catalog/queries/lookup/abr-cht.rq"
+        },
+        {
+          "@type": "Action",
+          "actionApplication": {
+            "@id": "https://reconciliation-api.github.io/specs/latest/",
+            "@type": "SoftwareApplication"
+          }
```
ddeboer, line 45: Does this make sense as a way to indicate that the Reconciliation API should include this dataset (distribution)? It’s a bit weird that this is an action on the SPARQL endpoint. On the other hand, the SPARQL endpoint is exactly what we’re using to fulfil reconciliation requests.

packages/network-of-terms-catalog/catalog/datasets/gtaa-classificatie.jsonld
```diff
@@ -30,6 +30,13 @@
         {
           "@type": "FindAction",
           "query": "file://catalog/queries/lookup/gtaa.rq"
+        },
+        {
+          "@type": "Action",
+          "actionApplication": {
+            "@id": "https://reconciliation-api.github.io/specs/latest/",
+            "@type": "SoftwareApplication"
+          }
```
ddeboer, line 39: @wmelder This will add support for Reconciliation Service for GTAA datasets in the Network of Terms. Please let me know your new GTAA infrastructure already supports the [Reconciliation Service API](https://reconciliation-api.github.io/specs/latest/). In that case the Network of Terms should refer to that instead.

wmelder: @ddeboer This is the first time that I see this service specification, so I don't expect this to be supported by the GTAA service layer.

ddeboer: To add some context: the Reconciliation spec is used (mostly) by OpenRefine to match strings against things (in our case, URIs of terms). 

No problem if GTAA doesn’t have this. In that case the Network of Terms provides this for you, so users can still reconcile their data with terms from the GTAA.

wmelder: ok, that's great!

packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
sdevalk: Yes, this makes sense to me. If a source, for whatever reason, provides two SPARQL distributions (i.e. in the Dataset Registry), we can use this model to denote which one must be used for recon.

I'm a little bit unsure about `actionApplication` as predicate of the `Action`. If I understand https://schema.org/actionApplication correctly, it should be used on an `EntryPoint`?

packages/network-of-terms-query/src/catalog.ts
```diff
@@ -58,3 +70,7 @@ export class SparqlDistribution {
 export type Distribution = SparqlDistribution;
 
 export class IRI extends URL {}
+
+export enum Feature {
+  RECONCILIATION = 'https://reconciliation-api.github.io/specs/latest/',
+}
```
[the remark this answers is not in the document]
sdevalk: Excellent!

sdevalk (commented):
Nice! Just one question :)

PR #540, opened 2022-03-04
feat: Score Reconciliation results based on distance

ddeboer:
* Score Reconciliation results by case-insensitive Levenshtein distance. E.g. when searching for ‘hout’:
  Screenshot 2022-03-04 at 11 20 00 
* If result has no prefLabels, default to score of 0.
* Sort results in order of decreasing scores, as recommended by the Reconciliation spec.


packages/network-of-terms-reconciliation/src/query.ts
```diff
@@ -36,19 +38,41 @@ export async function reconciliationQuery(
...
+  return (
+    1 /
+    term.prefLabels.reduce(
+      (distance, prefLabel) =>
+        distance +
+        leven(queryString.toLowerCase(), prefLabel.value.toLowerCase()),
+      1
+    )
+  );
+};
```
ddeboer, line 74: Do these scores make sense?

- identical word = score 1
- any differences decrease that score to fractions of 1, e.g. 0.35

The Reconciliation spec doesn’t specify that the scores must look like, but their examples seem to be on a score of 1–100(?). Which scores are better readable for humans? See also the OpenRefine screenshot above.

packages/network-of-terms-reconciliation/src/query.ts
```diff
@@ -36,19 +38,41 @@ export async function reconciliationQuery(
...
+ * prefLabels.
+ */
+const score = (queryString: string, term: Term): number => {
+  if (term.prefLabels.length === 0) {
+    return 0;
+  }
+
+  return (
+    1 /
+    term.prefLabels.reduce(
```
ddeboer, line 67: This only looks at prefLabels. Should we also look at other properties?

ddeboer:
Alternatively I tried scores in a range of 0–100 (percentage), as the Reconciliation spec’s examples suggest:

Screenshot 2022-03-04 at 12 27 33

Is this clearer? Apparently OpenRefine rounds to whole integers in this case, even though the API returns scores like `13.33`.

packages/network-of-terms-reconciliation/src/query.ts
sdevalk: Good one!

I don't have a strong opinion about this, but I'd prefer 1-100: it makes it easier to say that word 'hout' matches label 'hout' 100% and label 'Maashout' 50%.

packages/network-of-terms-reconciliation/src/query.ts
sdevalk: Interesting thought! In my sample dataset I have persons named 'Jan Havicksz. Steen', 'Maxime François Antoine Lalanne' and 'Herman van Dodeweerd'. If I reconcile these with e.g. RKDartists our Recon API returns terms such as:

https://data.rkd.nl/artists/74809
prefLabel: Steen, Jan
altLabel: Steen, Jan Havicksz.

https://data.rkd.nl/artists/47489
prefLabel: Lalanne, Maxime
altLabel: Lalanne, Maxime François Antoine

https://data.rkd.nl/artists/2458
prefLabel: Armando
altLabel: Dodeweerd, Herman Dirk van

In other words: perfect matches, because of the altLabels. If we only calculate the scores based on prefLabels, the perfect matches may get a lower ranking. So it makes sense to also look at the altLabels. Perhaps we could calculate a score for both the prefLabel and the altLabel and use the highest score in the result?

sdevalk: If that's a sound approach, we could also take the hiddenLabel into consideration.

sdevalk: If that's a sound approach, we could also take the hiddenLabel into consideration.

ddeboer: I have some hesitations about the hiddenLabel because that is not shown in the reconciliation results or preview, so it would become rather opaque to users what the score is based on.

packages/network-of-terms-reconciliation/src/query.ts
ddeboer: Yep, let’s stick with percentages.

packages/network-of-terms-reconciliation/src/query.ts
sdevalk: Yes, good point! On the other hand: that seems to be the exact application of `hiddenLabel`. From the SKOS primer [1]:

> A hidden lexical label, represented by means of the [skos:hiddenLabel](http://www.w3.org/TR/skos-reference#hiddenLabel) property, is a lexical label for a resource, where a KOS designer would like that character string to be accessible to applications performing text-based indexing and search operations, but would not like that label to be visible otherwise.

I could ask the Recon API testers what they think of it (i.e. should we look at hiddenLabel or not)?

[1] https://www.w3.org/TR/skos-primer/

ddeboer: Please do! Just to be clear: the current (released) functionality does _not_ look at hiddenLabel for scoring. 

But because it does consider hiddenLabel to determine _whether to include a term in the search results_ in the first place, we can expect a problem with larger result sets, when terms that match well on hiddenLabel (but do not on their prefLabel/altLabel) get scored 0 and drop out of the results together. This problem will become even clearer when clients set a relatively low limit (e.g. 3).

sdevalk: Thanks, very clear!

PR #559, opened 2022-03-14
optimized query for gtaa persoonsnamen

wmelder:
Query for the GTAA persoonsnamen can take quite a while because of the size of the term source (>200k). Therefore, the query was defined so that it would only allow single word source. Also, the number of result concepts was limited to 10. So I thought. It turned out the limit was on triples, because of the CONSTRUCT query type. Now the limit is raised to 50, so that it will return some 10 concepts when found. Still, whenever a term is searched that is very unique and 50 triples aren't produced, the whole dataset is being searched and timeouts can happen.

### DIFF
--- a/packages/network-of-terms-catalog/catalog/queries/search/gtaa-persoonsnamen.rq
+++ b/packages/network-of-terms-catalog/catalog/queries/search/gtaa-persoonsnamen.rq
@@ -1,55 +1,28 @@
 PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
 PREFIX justskos: <http://justskos.org/ns/core#>
-
 CONSTRUCT {
     ?uri a skos:Concept ;
         skos:prefLabel ?prefLabel ;
         skos:altLabel ?altLabel ;
         skos:hiddenLabel ?hiddenLabel ;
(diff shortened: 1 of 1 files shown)


packages/network-of-terms-catalog/catalog/queries/search/gtaa-persoonsnamen.rq
```diff
@@ -62,20 +35,6 @@ WHERE {
...
-    OPTIONAL {
-        ?uri skos:narrower ?narrower_uri .
-        ?narrower_uri skos:prefLabel ?narrower_prefLabel .
-        FILTER(LANG(?narrower_prefLabel) = "nl")
-    }
-    OPTIONAL {
-        ?uri skos:related ?related_uri .
-        ?related_uri skos:prefLabel ?related_prefLabel .
-        FILTER(LANG(?related_prefLabel) = "nl")
-    }
```
ddeboer, line 79: Did you remove broader, narrower and related because Persoonsnamen doesn’t have those relations or only to limit the number of triples returned per name in order to speed up the query?

ddeboer:
Thanks!

PR #581, opened 2022-03-24
feat: Show term source in reconciliation preview

### LINKED ISSUE #579: Show source on term preview card

ddeboer:
Fix #579.

### DIFF
--- a/packages/network-of-terms-reconciliation/src/preview.ts
+++ b/packages/network-of-terms-reconciliation/src/preview.ts
@@ -1,4 +1,5 @@
 import {
+  Dataset,
   LookupQueryResult,
   RelatedTerm,
   Term,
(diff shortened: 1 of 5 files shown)


packages/network-of-terms-reconciliation/src/preview.ts
```diff
@@ -31,6 +36,8 @@ export function preview(lookupResult: LookupQueryResult, locale: locale) {
         ${relatedTerms(locale.broader, term.broaderTerms)}
         ${relatedTerms(locale.narrower, term.narrowerTerms)}
         ${relatedTerms(locale.related, term.relatedTerms)}
+        <dt>${locale.source}</dt>
+        <dd>${source.name} (${source.creators[0]?.alternateName})</dd>
```
ddeboer, line 40: I decided it would be best to start the card with the term itself rather than its source. Does it make sense to show the term source like this, at the bottom of the card, in the definition list?

sdevalk: Perfect!

sdevalk (approved):

PR #582, opened 2022-03-24
feat: Improve reconciliation scoring

### LINKED ISSUE #580: Improve scoring
While scoring is already case-insensitive, we can improve it further by:

- ignoring word order (e.g. [`Pieter de Hooch` versus `Hooch, Pieter de`](https://termennetwerk.netwerkdigitaalerfgoed.nl/lookup?uri=https://data.rkd.nl/artists/39452))
- removing punctuation (`,` in this example).

ddeboer:
* Use cosine similarity instead of Levenshtein to ignore differences in
  word order.
* Ignore punctuation marks.

Fix #580.

### DIFF
--- a/packages/network-of-terms-reconciliation/test/server.test.ts
+++ b/packages/network-of-terms-reconciliation/test/server.test.ts
@@ -60,11 +60,11 @@ describe('Server', () => {
     // Results must be sorted by score in decreasing order.
     expect(results.q2.result).toHaveLength(2);
     expect(results.q2.result[0].name).toEqual('All things art');
-    expect(results.q2.result[0].score).toEqual(42.86); // Match of ‘things’ in prefLabel ‘All things art’.
+    expect(results.q2.result[0].score).toEqual(81.65); // Match of ‘things’ in prefLabel ‘All things art’.
     expect(results.q2.result[1].description).toEqual(
       'painted things that can be beautiful • another altLabel'
     ); // Result has no prefLabel.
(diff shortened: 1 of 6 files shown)


packages/network-of-terms-reconciliation/test/server.test.ts
```diff
@@ -60,11 +60,11 @@ describe('Server', () => {
     // Results must be sorted by score in decreasing order.
     expect(results.q2.result).toHaveLength(2);
     expect(results.q2.result[0].name).toEqual('All things art');
-    expect(results.q2.result[0].score).toEqual(42.86); // Match of ‘things’ in prefLabel ‘All things art’.
+    expect(results.q2.result[0].score).toEqual(81.65); // Match of ‘things’ in prefLabel ‘All things art’.
     expect(results.q2.result[1].description).toEqual(
       'painted things that can be beautiful • another altLabel'
     ); // Result has no prefLabel.
-    expect(results.q2.result[1].score).toEqual(16.67); // Match of ‘things’ in altLabel ‘painted things that can be beautiful’.
+    expect(results.q2.result[1].score).toEqual(63.25); // Match of ‘things’ in altLabel ‘painted things that can be beautiful’.
```
ddeboer, line 67: These are quite drastic changes in the scores: a word matching a single word in a much larger string is now scored higher.

sdevalk: Yes, indeed! However, I can see the benefits of these new scores: I'd like to give them a try in my test dataset. So: let's merge and find out! :)

sdevalk (approved):

PR #591, opened 2022-03-29
fix: Improve Wikidata streets search

ddeboer:
* ‘Schoolstraat Nijmegen’ returned no matches, whereas ‘Schoolstraat’ didn’t
  include Schoolstraat in Nijmegen due to the LIMIT on the results.
* Replace `mwapi:search` with `mwapi:srsearch`, which seems more capable.
* Simplify CONSTRUCT.
* Only construct altLabel manually if the result doesn’t already come with an
  an altLabel. This prevents duplicate altLabels ‘Street (City), Street (City)’.


EnnoMeijers (commented):
Nice clean up! I remember choosing the EntitySearch api for better performance as it only searches the titles and not the full text like the Search api does. But everything seems to be working fine this way now.

PR #592, opened 2022-03-30
feat: Add Iconclass terminology source

### LINKED ISSUE #512: Add Iconclass terminology source
@epoz is working on adding the [Iconclass classification system](https://test.iconclass.org) to the Network of Terms. We can track questions and progress here.

ddeboer:
* Add provisional Iconclass source: the SPARQL endpoint
  does not support fulltext search, so use `CONTAINS` instead.
* Combining this with `OPTIONAL` clauses slows down the query too much,
  so leave them out for now.
* Do not yet add Iconclass ‘no keys’ source because the semantics of that are 
  still unclear (see question in this PR).

Fix #512.


packages/network-of-terms-catalog/catalog/queries/search/iconclass.rq
```diff
@@ -0,0 +1,33 @@
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+
+CONSTRUCT {
```
ddeboer, line 3: CONSTRUCT queries are not yet supported.

packages/network-of-terms-catalog/catalog/datasets/iconclass-no-keys.jsonld
```diff
@@ -0,0 +1,37 @@
+{
+  "@context": "https://schema.org/docs/jsonldcontext.jsonld",
+  "@id": "https://test.iconclass.org",
+  "@type": "Dataset",
+  "name": [
+    {
+      "@language": "nl",
+      "@value": "Iconclass: geen keys"
```
ddeboer, line 8: How should we translate ‘keys’?

packages/network-of-terms-catalog/catalog/queries/search/iconclass.rq
```diff
@@ -0,0 +1,33 @@
...
+
+    ?broader_uri skos:prefLabel ?broader_prefLabel .
+    ?narrower_uri skos:prefLabel ?narrower_prefLabel .
+    ?related_uri skos:prefLabel ?related_prefLabel .
+}
+WHERE {
+    ?uri a skos:Concept ;
+        skos:prefLabel ?prefLabel .
+
+    ?uri <http://iconclass.org/search> ?query .
```
ddeboer, line 18: Using this predicate always returns zero search results.

PR #605, opened 2022-04-14
feat: Return feature URL from catalog

ddeboer:
BREAKING CHANGE: `dataset.distribution.features` is now a list of
`Feature`s that contain both the type and the URL at which the feature
is available.

### DIFF
--- a/packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
+++ b/packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
@@ -44,7 +44,8 @@
             "actionApplication": {
               "@id": "https://reconciliation-api.github.io/specs/latest/",
               "@type": "SoftwareApplication"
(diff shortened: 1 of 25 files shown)


packages/network-of-terms-catalog/catalog/datasets/abr.jsonld
```diff
@@ -44,7 +44,8 @@
             "actionApplication": {
               "@id": "https://reconciliation-api.github.io/specs/latest/",
               "@type": "SoftwareApplication"
-            }
+            },
+            "urlTemplate": "https://termennetwerk-api.netwerkdigitaalerfgoed.nl/reconcile/{distribution}"
```
ddeboer, line 48: I’m lazy so using a template here that’s the same for all reconciliation sources: only the distribution IRI is different.

PR #615, opened 2022-04-23
fix: Return correct source when sources share URI prefix

### LINKED ISSUE #492: GTAA lookup doesn’t work

### LINKED ISSUE #604: reconcilation preview shows the wrong source
When reconciling against the NTA the preview shows the "Brinkman Trefwoordenthesaurus" as the terminology source.

ddeboer:
* Queries can return a `skos:inScheme` triple that will be used
  instead of the dataset matching the URI prefix.

Fix #492.
Fix #604.

### DIFF
--- a/packages/network-of-terms-query/src/lookup/lookup.ts
+++ b/packages/network-of-terms-query/src/lookup/lookup.ts
@@ -48,14 +48,17 @@ export class LookupService {
     const irisToDataset = iris.reduce((acc, iri) => {
       const dataset = this.catalog.getDatasetByTermIri(iri);
       if (dataset) {
-        acc.set(iri, dataset);
+        acc.set(iri.toString(), dataset);
       }
       return acc;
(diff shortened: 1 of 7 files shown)


packages/network-of-terms-query/src/lookup/lookup.ts
```diff
@@ -69,15 +72,36 @@ export class LookupService {
...
+          }
+          termsResult.terms.push(term);
+        }
+        acc.set(dataset, termsResult);
+      } else {
+        const dataset = this.catalog.getDatasetByDistributionIri(
+          result.distribution.iri
+        )!;
+        acc.set(dataset, result);
+      }
```
ddeboer, line 99: This added complexity is anything but elegant, so we may want to refactor later.

PR #647, opened 2022-05-13
fix: STCN

### LINKED ISSUE #500: Extend STCN Drukkers information
For the STCN Drukkers thesaurus new entities are defined when a printer moves from one place to another. For example printer http://data.bibliotheken.nl/doc/thes/p075573636 refers to Plantijn working in Antwerp and http://data.bibliotheken.nl/doc/thes/p075556251 refers to Plantijn working in Leiden. In the data this is expressed through the schema:addressLocality field. In the results the difference between both entities is not visibile. My suggestions is to add schema:addressLocality (if available) to the skos:scopeNote stating the locality of the printer.

EnnoMeijers:
Added location info to lookup and search query, this fixes #500. We should also add time periods in which printers were active but these are not yet available in the Linked Data.

### DIFF
--- a/packages/network-of-terms-catalog/catalog/queries/search/stcn-drukkers.rq
+++ b/packages/network-of-terms-catalog/catalog/queries/search/stcn-drukkers.rq
@@ -25,9 +25,8 @@ WHERE {
     # http://data.bibliotheken.nl/id/thes/p338012834 for Plantijn working in Antwerp
     # The original source (STCN) shows also a time period, this data currently not available in the LOD version.
     OPTIONAL { 
-     ?uri schema:location/schema:address/schema:addressLocality ?place 
+     ?uri schema:location/schema:address/schema:addressLocality ?scopeNote 
     }
-    BIND(COALESCE(IF(BOUND(?place), ?place, ?noPlace)) as ?scopeNote)
     OPTIONAL { ?uri schema:name ?schema_name }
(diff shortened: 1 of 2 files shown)


packages/network-of-terms-catalog/catalog/queries/search/stcn-drukkers.rq
```diff
@@ -16,9 +18,17 @@ WHERE {
...
+    # STCN defines multiple URIs for the same printer.
+    # The distinction between these URIs is based on the place where the printer lived and worked.
+    # See for example "Plantijn"
+    # http://data.bibliotheken.nl/id/thes/p075556251 for Plantijn working in Leiden
+    # http://data.bibliotheken.nl/id/thes/p338012834 for Plantijn working in Antwerp
+    # The original source (STCN) shows also a time period, this data currently not available in the LOD version.
+    OPTIONAL { 
+     ?uri schema:location/schema:address/schema:addressLocality ?place 
+    }
+    BIND(COALESCE(IF(BOUND(?place), ?place, ?noPlace)) as ?scopeNote)
```
ddeboer, line 30: Couldn’t this be simplified to:

```suggestion
    OPTIONAL { 
        ?uri schema:location/schema:address/schema:addressLocality ?scopeNote
    }
```

EnnoMeijers: Good point, a left over of a more complex approach, removed the BIND statement

PR #684, opened 2022-06-21
Geonames

EnnoMeijers:
Added geonames queries and datasource definition


EnnoMeijers:
ok, thanks! I am kind of lost here at the moment....

PR #685, opened 2022-06-22
Feat: added geonames

EnnoMeijers:
Started new geonames branch to resolve merging conflicts.

### DIFF
--- a/packages/network-of-terms-catalog/catalog/queries/search/geonames.rq
+++ b/packages/network-of-terms-catalog/catalog/queries/search/geonames.rq
@@ -1,0 +1,45 @@
+PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
+PREFIX rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#>
+PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
+PREFIX gn: <https://www.geonames.org/ontology#>
+
+CONSTRUCT {
+    ?uri a skos:Concept ;
+        skos:prefLabel ?prefLabel ;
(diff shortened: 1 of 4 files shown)


packages/network-of-terms-catalog/catalog/queries/search/geonames.rq
```diff
@@ -0,0 +1,45 @@
...
+    ?uri skos:broader ?adm2_uri .
+    ?adm2_uri skos:prefLabel ?adm2_prefLabel .
+}
+WHERE { 
+  # search in names and alternative names
+  ?uri ?predicate ?name .
+  VALUES ?predicate {gn:name gn:alternateName } .
+  FILTER(CONTAINS(LCASE(?name),LCASE($query)))
+  
+  # limit results to places (P), localities (L), administrative levels (A) and water surfaces (H)
```
ddeboer, line 22: Why should we filter by feature class? Some random result, https://www.geonames.org/2802421/beerzelschrans.html (a farm) turns out to be interesting from a cultural heritage perspective: https://inventaris.onroerenderfgoed.be/erfgoedobjecten/2461.

EnnoMeijers: My suggestion is to start small and enlarge the selection on requests from users. I see your point in the farm example, on the other hand geonames also contains a lot of other entries like hotels https://www.geonames.org/6466448/golden-tulip-westduin-vlissingen-koudekerke.html which will pollute the result lists if we do not filter these. For now I have chosen a basic set of features as a start which I think will be useful for the average use case.

ddeboer (approved):
Made some changes, good to go now.

PR #767, opened 2022-10-17
fulltext querying enable in jena to increase preformance

wmelder:
- performance is improved, for large GTAA conceptschemes like Persoonsnamen.
- diacritics are handled
- separate query `gtaa-persoonsname.rq` can be removed.

PR #780, opened 2022-10-24
#change gtaa query inlcude hidden labels

wmelder:
as was signaled "zelensky" didn't produce results, while "zelenski" does.
Both should produce results, because "zelensky" is definitely in one of the hiddenLabels.
As it turns out, the default full text index doesn't include the hiddenlabels, therefor we changed the text query a bit..

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - name: Dependabot metadata
         id: metadata
-        uses: dependabot/fetch-metadata@v1.3.3
+        uses: dependabot/fetch-metadata@v1.3.4
         with:
           github-token: "${{ secrets.GITHUB_TOKEN }}"
       - name: Enable auto-merge for Dependabot PRs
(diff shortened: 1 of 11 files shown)


ddeboer:
Thanks!

ddeboer:
@wmelder However, this seems to cause [timeouts for GTAA persoonsnamen](https://termennetwerk.netwerkdigitaalerfgoed.nl/?q=zelenski&datasets=https://data.beeldengeluid.nl/id/datadownload/0026).

wmelder:
@ddeboer The timeouts should be fixed because of the text query. But indeed, the termennetwerk demonstrator does seem to give a terminology source unavailable. Will look into it.

wmelder:
@ddeboer please confirm that the gtaa-persoonsnamen.rq query is not used anymore. It should be replaced with the gtaa.rq, which uses textQuery and is fast enough in itself to search even large conceptschemes like Persoonsnamen.

ddeboer:
> @ddeboer please confirm that the gtaa-persoonsnamen.rq query is not used anymore. It should be replaced with the gtaa.rq, which uses textQuery and is fast enough in itself to search even large conceptschemes like Persoonsnamen.

It is still used: 

https://github.com/netwerk-digitaal-erfgoed/network-of-terms/blob/e32df4b8b3d1228222da01bc416ed24084ea0a9c/packages/network-of-terms-catalog/catalog/datasets/gtaa-persoonsnamen.jsonld#L34

Shouldn’t it be?

wmelder:
no, that is still the old query that doesn't use the fulltext functionality. 
All GTAA conceptSchemes can now use the same search query

ddeboer:
Ah, you mentioned so in #767 but didn’t actually commit the required changes. I can do so now.

PR #783, opened 2022-10-29
Feat: geonames update

EnnoMeijers:
Changed the description for the Geonames source to show the German data is now included. Also changed to search query to use STRSTARTS in stead of CONTAINS to reduce the noise.


packages/network-of-terms-catalog/catalog/queries/search/geonames.rq
```diff
@@ -24,7 +24,9 @@ WHERE {
     # Limit results to places (P), localities (L), administrative levels (A) and water surfaces (H).
     VALUES ?featureClass { gn:P gn:L gn:A gn:H }
 
-    FILTER(CONTAINS(LCASE(?name), LCASE($query)))
+    # Select results that start with the search string
+    # Using contains() leads too much noise
```
ddeboer, line 28:
```suggestion
    # Using contains() leads to too much noise 
```

PR #828, opened 2022-12-01
replace justskos namespace

wmelder:
Search on GTAA sources is not working due to a change on the triple store. 
To fix this: Small change that replaces the justskos prefix with openskos.


wmelder:
@ddeboer please merge PR because the search isn't working right now.

ddeboer:
Thanks @wmelder!

Issue #537, opened 2022-03-04
Rank results by full text score from sources


ddeboer:
Investigate whether we can return matching scores from sources that support them to increase result relevance for users.

Previously, we decided to sort results alphabetically because we cannot guarantee the order of results (#68). However, if the source supports a score property, we can sort on that, even client-side if necessary.

* For Virtuoso, this should be `?o <bif:contains> "gogh" OPTION (score ?sc)`, but see https://github.com/comunica/comunica/issues/940.
* Wikidata’s `mwapi` [doesn’t seem to support scores](https://www.mediawiki.org/wiki/Wikidata_Query_Service/User_Manual/MWAPI#Examples).
* …

ddeboer:
@wmelder Can you paste an example GTAA query here that returns relevance scores and/or orders the results by them?

ddeboer:
Thanks @wmelder, that was helpful. Results for GTAA are now [sorted by relevance](https://termennetwerk.netwerkdigitaalerfgoed.nl/?q=Joan+miro&datasets=https://data.beeldengeluid.nl/id/datadownload/0026) (and alphabetically for results with identical scores).

PR #876, opened 2023-02-01
replaced namespace and decreased limit

wmelder:
The namespace for the B&G properties is now published and the queries need to use it to work.

Also, the LIMIT was still on 1000, which seemed appropriate when the query seemed to limit the number of triples in the construct query. Now we have changed the query to use the full text capabilities, the LIMIT seems to be the number of concepts. Note that the Jena results are not ranked, but when the limit is set to 10, the 10 most relevant results are returned (although unordered). When using a limit 1000 the best result may end up on place 1000 in the worst case.

We use the Apache Lucene full text engine on Apache Jena triple store. This is able to send relevance information, which can be used to sort the results. Currently, the network of terms is not using such functionality. We could provide example queries and results that contain the relevance scores, so that the result list can be sorted. Please send a message when interested. For now we just suggest to bring down the number of results drastically, because that will improve the search.

### DIFF
--- a/packages/network-of-terms-catalog/catalog/queries/search/gtaa.rq
+++ b/packages/network-of-terms-catalog/catalog/queries/search/gtaa.rq
@@ -1,5 +1,5 @@
 PREFIX skos: <http://www.w3.org/2004/02/skos/core#>
-PREFIX openskos:  <http://openskos.org/xmlns#>
+PREFIX bengthes: <http://data.beeldengeluid.nl/schema/thes#>
 PREFIX text: <http://jena.apache.org/text#>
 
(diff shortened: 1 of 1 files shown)


wmelder:
@ddeboer Note that the search for GTAA terms in the network of terms is currently not working, due to the updated namespace. This was deployed on the SPARQL endpoint without notification beforehand, therefor I couldn't prepare a PR earlier. So, if you have a little time, please have look at this small PR. You can ignore the comments about sorting and ranking for now.

ddeboer:
Thanks @wmelder! Let’s continue the discussion about returning relevance scores in #537.

PR #877, opened 2023-02-06
feat: Sort by relevance score if available

### LINKED ISSUE #537: Rank results by full text score from sources

ddeboer:
* Queries may sort by relevance score.
* If they do, sort by score primarily. For identical scores, fall back to alphabetical sort.
* If they do not, only sort alphabetically.
* This PR does not yet expose the score in the GraphQL API.

Fix #537

### DIFF
--- a/packages/network-of-terms-query/src/terms.ts
+++ b/packages/network-of-terms-query/src/terms.ts
@@ -12,7 +12,8 @@ export class Term {
     readonly broaderTerms: RelatedTerm[],
     readonly narrowerTerms: RelatedTerm[],
     readonly relatedTerms: RelatedTerm[],
-    readonly datasetIri: RDF.Term | undefined
+    readonly datasetIri: RDF.Term | undefined,
+    readonly score: RDF.Literal | undefined
   ) {}
 }
 
@@ -32,6 +33,7 @@ class SparqlResultTerm {
   narrowerTerms: RDF.Term[] = [];
   relatedTerms: RDF.Term[] = [];
   inScheme: RDF.Term | undefined = undefined;
+  score: RDF.Literal | undefined = undefined;
 }
 
 export class TermsTransformer {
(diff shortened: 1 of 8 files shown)


packages/network-of-terms-query/src/terms.ts
```diff
@@ -55,6 +57,7 @@ export class TermsTransformer {
     ['http://www.w3.org/2004/02/skos/core#related', 'relatedTerms'],
     ['http://www.w3.org/2008/05/skos#related', 'relatedTerms'],
     ['http://www.w3.org/2004/02/skos/core#inScheme', 'inScheme'],
+    ['http://jena.apache.org/text#score', 'score'],
```
ddeboer, line 60: Of course, this is not the namespace that we want to use for storing the score in queries. It should be decoupled from any particular SPARQL store.

@EnnoMeijers Do you have ideas for a generic RDF vocabulary concerning search results and their relevance scores?

EnnoMeijers: Hmm, good question! Maybe use http://purl.org/voc/vrank, it seems to be around for quite some time, the vrank:rankValue property could be appropriate? Vrank has also properties to describe the ranking method in more detail.

ddeboer: Good suggestion, thanks!

Issue #913, opened 2023-03-01
Wikidata requests can result in timeouts due to use of HTTP URI


EnnoMeijers:
Wikidata seems to have a problem with resolving HTTP URIs to the HTTPS were the actual data is found. This was reported by Reinier van de Giessen using the TN API as implemented in their Atlantis system. A bug report to Wikidata was submitted see https://phabricator.wikimedia.org/T330906. Their advise is to use the HTTPS version of entity identifiers to be safe. This is problematic in general because all linked data entities in WD are HTTP based. A workaround for now would be to replace 'HTTP://' with 'HTTPS://' in the ?item var in all our Wikidata queries.

EnnoMeijers:
As decided in PR #914 instead of changing the code we will advise the users of the TN API to mint WD into HTTPS URIs before reusing them in order to prevent problems with accessing the HTTP version.

PR #914, opened 2023-03-01
Fix: WD queries returning https uris

EnnoMeijers:
Fixes for the WD queries to resolve issue #913


EnnoMeijers:
As pointed out in issue #914 WD community advises to use only HTTPS URIs and move away from using HTTP. At the same all the WD Linked Data stuff is based on HTTP URIs so there will be a major problem for using WD as a source for Linked Data when they step away from supporting redirection from HTTP to HTTPS.  It looks like the current problems are at regional and technical level and not the result of a change of policy. So I agree with your point of view that we should not fix this problem on this level as it will introduce a mix of linked data WD URIs with HTTP and HTTPS. I will close the PR for now.

PR #937, opened 2023-04-05
Update SPARQL lookup and search query for brabantse-gebouwen

kiivihal:
refactor(brabantse-gebouwen.rq): replace skos:altLabel with skos:hiddenLabel and simplify skos:scopeNote

refactor(brabantse-gebouwen.rq): remove buildingTypeScopeNote and addressScopeNote, add hiddenLabel and scopeNote to CONSTRUCT

The queries have been simplified because the information is now available in the source endpoint.


ddeboer:
Great! Thank you @kiivihal.

PR #943, opened 2023-04-16
feat: Use subqueries for better LIMIT

ddeboer:


EnnoMeijers:
Is still waiting for my approval? Do we feel confident enough about this change? Has this been tested enough already, it has an impact on many queries, should we do a step by step roll out maybe? We also need to look at recently added sources.

ddeboer:
First changes done in #1367.

ddeboer:
Replaced by #1479.

PR #1046, opened 2023-07-26
Fixed and extended docker compose setup

EnnoMeijers:
The current docker-compose.yml failed to run. Fixed this by adding a user 'node' instead of running as 'root'. Also brought the port mapping to the command line as the 'run' does not read them from the config. To ease up thing I created two flavors of the docker-compose including all the necessary commands, so there is one to run the GraphQL API and one for the Reconcilation API. Updated the doc accordingly.


ddeboer:
What exactly didn’t work? My guess is that 4fdeebfb00c997512968fad6611dd8df66d7916a is a sufficient fix.

docs/docker.md
```diff
@@ -1,31 +1,28 @@
...
-    /app # cd packages/network-of-terms-graphql
-    /app/packages/network-of-terms-graphql # npm run dev
+To run the Reconciliaton API run the development container using the following command:
 
-    # or:
+    docker compose -f docker-compose-reconciliation.yml run -p 8080:3123 --rm reconciliation
 
-    /app # cd packages/network-of-terms-reconciliation
-    /app/packages/network-of-terms-reconciliation # npm run dev
+Note that you need to restart the docker compose after every change in the code.
```
ddeboer, line 21: That’s too cumbersome for developers, which is why I suggested `npm run dev` (which watches for changes) before.

EnnoMeijers:
I added 'user: node' (or what ever user you like) to the docker-compose.yml to prevent the following error when running 'npm run dev':
```
glob error [Error: EACCES: permission denied, scandir '/root/.npm/_logs'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'scandir',
  path: '/root/.npm/_logs'
}
npm WARN logfile Error: EACCES: permission denied, scandir '/root/.npm/_logs'
```
If missed the fact that 'npm run dev' is wachting for changes (see your previous comment) so I created separate docker-compose files that take care of installing en starting the apps, it would be to tedious to do this by hand for every change. But ok to keep the original setup if that has your preference. 

The advantage of bringing the port mapping outside instead of using '--service-ports' is that you run both services at the same time on the port of your choice.

EnnoMeijers:
Closing this PR

Issue #1048, opened 2023-07-28
running tests in docker dev environment fails


EnnoMeijers:
When using the docker-compose.yml for running a local development instance of the NoT a number of tests fail probably due to misconfiguration of the internal network in the container.  

Reproduce with:
```
docker compose run --service-ports --rm node
npm install
npm run test
```
Needless to say that being able to run the tests is an important feature for a development environment.

BTW: all the tests pass when run them directly in a node environment.

ddeboer:
Can you post the full test output? The run fine on my machine.

EnnoMeijers:
See the file below:
[test-output.log](https://github.com/netwerk-digitaal-erfgoed/network-of-terms/files/12193319/test-output.log)

ddeboer:
* You’re getting `EAI_AGAIN`, which is a DNS error, that I don’t see on my machine.
* Your NPM version is 8.19.3, while mine is 9.x.

Can you try: `docker compose pull` to update your `node:lts-alpine` image and then start the container and run the tests again?

EnnoMeijers:
Did the pull and now it is running with npm version 9.6.7 but still the same result. Looks like an alpine related issue according to Google search. Or could it be docker related, my version is 20.10.24. Tried pruning my docker networks but with no result. I tried another (Ubuntu) laptop, docker version 23.0.5, same result :(

EnnoMeijers:
Using a different node images solved my problems, tried *slim versions but ran into different problems. Falling back to the default node image (node:18) seems to be the best solution for now. Created PR #1050

PR #1050, opened 2023-07-29
Changed node image

EnnoMeijers:
Switch from alpine to default node image to solve issue #1048

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   node:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: "3.9"
 services:
   node:
-    image: node:lts-alpine
+    image: node:18
```
ddeboer, line 4:
```suggestion
    image: node:lts
```

PR #1065, opened 2023-08-08
improve full text search for Geonames

EnnoMeijers:
Used `?virtuosoQuery` for Geonames query to improve support for search with phrases like `'gent sas van'.` Should not have any impact on single term searches, did some testing to confirm this. The `?virtuosoQuery` seems not to go well with the more complex `text:query` options using '`( ... )'`, had to remove this from the query but seemed redundant anyway.


ddeboer:
Good to know this works. However, particularly because the GeoNames SPARQL endpoint is managed by us, I prefer to solve this issue on the endpoint side.

EnnoMeijers:
Ok, I agree with that direction but if that needs a longer period of time to solve it I would strongly advise to do this update as a temporarily fix. The current multi word search is horrible, try searching for 'Den Haag'....

Issue #1064, opened 2023-08-07
handling greedy full text indexes


EnnoMeijers:
I ran into a case were the full text index is extremely greedy, see https://api.triplydb.com/s/45wCpVY7r .  The only way to work with this set is by using and AND clause on the searchs terms like this https://api.triplydb.com/s/KVXEVl14Y. 

Using queryMode=RAW in the GraphQL API enables this way of querying but it this is no option when using the Reconciliation Service API or the Demonstrator.  Could this be a default setting in the dataset definition? For reconcilation it should expand the search query with an AND clause when multiple terms are used. 

Any ideas how to handle this?

ddeboer:
> it should expand the search query with an AND clause when multiple terms are used.

We [already do so](https://github.com/netwerk-digitaal-erfgoed/network-of-terms/blob/1dcee6711dcfb588e3ee026ba4a3c348ef71e3de/packages/network-of-terms-query/src/search/query-mode.ts#L43-L50) for `?virtuosoQuery`; you can try using that in your SPARQL query from the NoT, although the backend seems to be Fuseki rather than Virtuoso. 

We could also add a third query variable besides `?query` and `?virtuosoQuery`. Something like `?intersectionQuery`, which joins all query words with `AND` but does not quoting like `?virtuosoQuery` does.

EnnoMeijers:
I think it would indeed be very useful to add to add another variable like that. Treating the phrase as "OR"ed words seems to be default  behaviour of the [Jena Full Text Search](https://jena.apache.org/documentation/query/text-query.html). We have a similar problem with the current Geonames implementation. Searching for ["Sas van Gent"](https://jena.apache.org/documentation/query/text-query.html) returns 195 results, all most all of them are irrelevant. Rewriting the query to `"(sas AND gent AND van)" `returns the two relevant ones, see https://api.triplydb.com/s/8DBt_82Qy.

Note: The GTAA also uses the Jena Full Text Search but does not have this problem, searching for [Wim de Bie](https://termennetwerk.netwerkdigitaalerfgoed.nl/?q=wim+de+bie&datasets=https://data.beeldengeluid.nl/id/datadownload/0026) returns the correct results. Maybe we should approach @wmelder to asks for details on his config settings?

rschalkrce:
Thanks for signalling this @EnnoMeijers. I agree, using an AND clause for multiple terms should preferrably be standard. 

[...]

PR #1066, opened 2023-08-08
feat: Sort GeoNames by relevance score

ddeboer:
Ref #1064
Replace #1065


wmelder:
Question: you're using `vrank:simpleRank` on an instance of a `skos:Concept`. The formal definition of the property describes it should be an `owl:Thing` as the domain:
```turtle
vrank:simpleRank a rdf:Property, owl:DatatypeProperty, owl:FunctionalProperty;
    owl:propertyChainAxiom (vrank:hasRank vrank:rankValue) ;
    rdfs:label "rank";
    rdfs:comment "Directly connects owl:Thing with ranking values.";
    rdfs:domain owl:Thing;
    rdfs:range xsd:float .
```
How are we allowed to still use it?

ddeboer:
Good question. As far as I can see, `owl:Thing` represents the set of all individuals, making all user-defined `owl:Class`es, including `skos:Concept` subclasses of `owl:Thing`. The domain `owl:Thing` then includes `skos:Concept`.

PR #1147, opened 2023-10-28
feat!: Add link to dataset page

ddeboer:
BREAKING CHANGE: for custom catalogs, add a `schema:mainEntityOfPage`
property to each dataset.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -11,9 +11,9 @@ export default {
   ],
   coverageThreshold: {
     global: {
-      lines: 90.89,
-      statements: 90.89,
-      branches: 95.89,
+      lines: 90.94,
+      statements: 90.94,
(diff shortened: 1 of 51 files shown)


ddeboer:
@rschalkrce I used [the property `schema:mainEntityOfPage`](https://netwerk-digitaal-erfgoed.github.io/requirements-datasets/#dataset-overview). Please check all my suggestions in the catalog, paying particular attention to the `https://???` values that I could not find any good webpage for.

rschalkrce (changes requested):
@ddeboer ,

I've added a URL to `mainEntityOfPage` where I could find one. Some are still open. For these I would suggest to continue to use the `url` value for now (I have not done that yet).

PR #1186, opened 2023-12-14
feat: Add Rijksmonumenten

rschalkrce:


ddeboer:
Hey @rschalkrce, very cool that you’re adding this yourself! 

You made three separate pull requests (PR), which on the one hand is good practice (small, specific PRs), but in the case of adding sources we usually combine adding the source and its queries in a single PR. That’s also needed to make the tests pass, because this PR does not contain the query files referenced the description.

Let me know if you want to try to merge all changes into this `rschalkrce-rijksmonumenten` branch yourself or would like me to do so. It’s fine with me either way.

rschalkrce:
Hi @ddeboer 

Thanks! I figured one PR would have been more appropriate, but could not see how to do that quickly using the browser (next time i'll use Github CLI/Desktop). 

If you can merge all changes into `rschalkrce-rijksmonumenten` that would be great!

packages/network-of-terms-catalog/catalog/datasets/rijksmonumenten.jsonld
```diff
@@ -0,0 +1,64 @@
...
+      {
+        "@id": "https://www.cultureelerfgoed.nl"
+      }
+    ],
+    "url": [
+      "https://linkeddata.cultureelerfgoed.nl/cho-kennis/id/rijksmonument/"
+    ],
+    "mainEntityOfPage": [
+      "https://linkeddata.cultureelerfgoed.nl/rce/Rijksmonument"
+    ],
```
ddeboer, line 27: @rschalkrce I had to add this property to make the description valid. Please let me know if a better landing page for this dataset is available.

packages/network-of-terms-catalog/catalog/datasets/rijksmonumenten.jsonld
```diff
@@ -0,0 +1,64 @@
...
+    "description": [
+      {
+        "@language": "nl",
+        "@value": "Rijksmonumenten"
+      }
+    ],
+    "inLanguage": "nl",
```
ddeboer, line 34: Added this for the same reason. Is this correct?

ddeboer:
So this is now live @rschalkrce.

Just wondering: this now shows

> Rijksmonumenten (Rijksmonumentenregister)

in the dropdown on https://termennetwerk.netwerkdigitaalerfgoed.nl.

Is that intended? Usually the `alternateName` between parentheses is a shorter name or abbreviation, such as AAT or CHT. Should we invert the two? And perhaps remove `alternateName` is there’s no common abbreviation for the source?

rschalkrce:
@ddeboer let's indeed just remove the alternate name as. there is no abbreviation here

Issue #1424, opened 2024-09-16
Lookup can't find URI's of streets in Gouda / always disappeared streets


coret:
When I lookup a street in Gouda, eg. https://termennetwerk.netwerkdigitaalerfgoed.nl/?q=scheygrondplantsoen&datasets=https://www.goudatijdmachine.nl/id/straten%23streets I get an URI like https://n2t.net/ark:/60537/b9MVd1
But when I use the lookup function, eg. https://termennetwerk.netwerkdigitaalerfgoed.nl/lookup?uri=https://n2t.net/ark:/60537/b9MVd1, I get a "Geen terminologiebron gevonden voor deze URI" message.

Additionally: all streets are called "verdwenen straten", which is incorrect.

goudatijdmachine-straten.rq's of search and lookup should be fixed

PR #1127, opened 2023-10-06
Gouda Tijdmachine query update

coret:
Adjusted queries to use the new GTM ontology (more schema.org) and to make use of a Lucene-index.

PR #1425, opened 2024-09-16
Added missing prefix and replaced non-existing ?entity to fix search & lookup of goudatijdmachine-straten

coret:
Fix for [#1424](https://github.com/netwerk-digitaal-erfgoed/network-of-terms/issues/1424)

Strange that these SPARQL errors weren't noticed by the NoT software.


ddeboer:
> Strange that these SPARQL errors weren't noticed by the NoT software.

@coret You added this yourself in https://github.com/netwerk-digitaal-erfgoed/network-of-terms/pull/1127.

The error was properly [logged](https://statistieken.netwerkdigitaalerfgoed.nl/d/e191c3e0-7d4e-4c1a-bc6a-171ef0b1fdf2/logs?orgId=1&var-app=network-of-terms-graphql&var-app=network-of-terms-reconciliation&var-filter=omeka&from=now-12h&to=now&var-stream=All), so not sure what you expect here.

PR #1426, opened 2024-09-17
feat: Identify datasets by their URI

### LINKED ISSUE #1423: Use dataset URI instead of distribution URI
Previously (#17) we decided to use the distribution URI to identity terminology sources to NoT clients. We’re running in some problems, for example:

- The RKD distribution URI was [changed](https://github.com/netwerk-digitaal-erfgoed/network-of-terms/pull/1409/files#diff-a32a33e9c1f1bc323a80fc499a083dcf630316e7d063f1df1d1fd80d3a6126efR36), which has caused trouble for some suppliers (Atlantis).
- Same for NMVW (https://github.com/netwerk-digitaal-erfgoed/network-of-terms/pull/1411).

We could support both dataset and distribution URI first, and possibly phase out distribution URIs as identifiers in the future.

ddeboer:
Fix #1423

Terminology sources are no longer identifier by their distribution URI.
Keep BC by still allowing the distribution URI to be used when searching.
The sources query now returns the new source instead of distribution URIs.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -12,7 +12,7 @@ export default {
     global: {
       lines: 91.4,
       statements: 91.4,
-      branches: 95.45,
+      branches: 95.46,
       functions: 92.62,
     },
   },
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,10 +10,10 @@ export default {
   ],
   coverageThreshold: {
     global: {
-      lines: 91.37,
-      statements: 91.37,
-      branches: 95.08,
-      functions: 92.56,
+      lines: 91.39,
+      statements: 91.39,
+      branches: 95.45,
+      functions: 92.62,
     },
   },
   transform: {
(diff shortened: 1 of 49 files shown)


EnnoMeijers (commented):
Looks good to me but we should ask the PO (Ruben) for explicite approval as it a quite a fundamental change.

rschalkrce:
As long as we continue both distribution and source ID for the time being it looks good to me as well. In the longer run we should communicate the move to dataset ID's first before phasing out distribution ID's. Just to be sure, this PR does not result in a breaking change, right?

ddeboer:
> Just to be sure, this PR does not result in a breaking change, right?

There’s no breaking change for clients when they send queries as they can continue using the distribution URI as before.

However, there _may_ be a breaking change for clients when reading the results, if they compare the URI from the results (which is now the dataset URI) with the distribution URI they may have configured locally before (e.g. in a collection management system): these no longer match.

PR #1431, opened 2024-09-23
fix: Reconciliation API URLs

ddeboer:


packages/network-of-terms-catalog/test/catalog.test.ts
```diff
@@ -84,14 +89,8 @@ describe('Catalog', () => {
     const reconciliationApis = catalog.getDistributionsProvidingFeature(
       FeatureType.RECONCILIATION
     );
-    expect(reconciliationApis[0].features).toContainEqual(
-      new Feature(
-        FeatureType.RECONCILIATION,
-        new URL(
-          'https://termennetwerk-api.netwerkdigitaalerfgoed.nl/reconcile/' +
-            reconciliationApis[0].iri.toString().replace('#', '%23')
```
ddeboer, line 92: This didn’t work, so move this assertion to the test above.

PR #1444, opened 2024-10-02
fix: goudse straten - pid property change

coret:
PIDs in the GTM KG used to be in the sdo:identifier property, but have moved to the owl:sameAs property, this made the ETL pipeline much easier.


ddeboer:
Should’t the lookup query be changed, too?

coret:
No, because the PIDs of the terms stay the same (and the magic of owl:sameAs).

PR #1446, opened 2024-10-28
Change ainEntityOfPage for Muziekschatten and Uitvoeringsmedium

rschalkrce:
Asked by data owner to change the `mainEntityOfPage` for these three datasets

### DIFF
--- a/packages/network-of-terms-catalog/catalog/datasets/muziekschatten-onderwerpen.jsonld
+++ b/packages/network-of-terms-catalog/catalog/datasets/muziekschatten-onderwerpen.jsonld
@@ -25,7 +25,7 @@
     "https://data.muziekschatten.nl/som/"
   ],
   "mainEntityOfPage": [
-    "https://data.muziekschatten.nl/som/"
+    "https://data.muziekschatten.nl/"
   ],
   "description": [
     {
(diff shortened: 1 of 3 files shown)


packages/network-of-terms-catalog/catalog/datasets/muziekschatten-onderwerpen.jsonld
```diff
@@ -22,7 +22,7 @@
     }
   ],
   "url": [
-    "https://data.muziekschatten.nl/som/"
+    "https://data.muziekschatten.nl/"
```
ddeboer, line 25: Red alert: does this mean URIs have changed? Or just that _new_ concepts may not have `/som/` in their URL?

packages/network-of-terms-catalog/catalog/datasets/uitvoeringsmedium.jsonld
```diff
@@ -22,7 +22,7 @@
     }
   ],
   "url": [
-    "https://data.muziekschatten.nl/som/um"
+    "https://data.muziekschatten.nl/som/"
```
ddeboer, line 25: So Uitvoeringsmedium still has `/som/` but subjects do not?

rschalkrce:
Woops, my mistake! It should only concern `MainEntityOf`, so the dataset landing pages for Muziekschatten and Uitvoeringsmedium.

ddeboer:
Fixed.

PR #1448, opened 2024-10-28
feat: Add Thesaurus Historische Persoonsgegevens

ddeboer:
* Re-use Poolparty query and adjust for `skos:definition`,
  used by PiCoT.


packages/network-of-terms-catalog/catalog/datasets/picot.jsonld
```diff
@@ -0,0 +1,69 @@
...
+  "name": [
+    {
+      "@language": "nl",
+      "@value": "Thesaurus Historische Persoonsgegevens"
+    }
+  ],
+  "alternateName": [
+    {
+      "@language": "en",
+      "@value": "PiCoT"
```
ddeboer, line 14: @RubenSchalk @pwoltjer This is the (common) abbreviation for this source; cf. NTA, GTAA etc. I changed this from `PiCo` to `PicOT` to differentiate the _ontology_ from the _terms list_. I saw @pwoltjer (👋) using `picot:` in the [examples](https://github.com/CBG-Centrum-voor-familiegeschiedenis/PiCo/blob/main/examples/extending-pico/slavenregister.ttl#L7), so `PiCoT` seems to make sense. Agreed?

packages/network-of-terms-catalog/catalog/datasets/picot.jsonld
```diff
@@ -0,0 +1,69 @@
...
+  "url": [
+    "https://terms.personsincontext.org/ThesaurusHistorischePersoonsgegevens/"
+  ],
+  "mainEntityOfPage": [
+    "https://terms.personsincontext.org/ThesaurusHistorischePersoonsgegevens"
+  ],
+  "description": [
+    {
+      "@language": "nl",
+      "@value": "Rollen voor het beschrijven van personen in historische documenten"
```
ddeboer, line 36: @rschalkrce @pwoltjer Not just roles but also source types, though. Can you provide a better description?

rschalkrce: I will pass this on to Pieter

PR #1615, opened 2025-09-09
Patch friese thesaurus personen see also prop

Xentropics:
As the Frisian Person Thesaurus URIs now resolve to an actual human readible page, there's no need to send users off to one of the mainEntityOfPage URIs anymore.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -10,10 +10,10 @@ export default {
   ],
   coverageThreshold: {
     global: {
-      statements: 92.34,
-      branches: 93.62,
-      lines: 92.34,
-      functions: 91.66,
+      statements: 92.5,
+      branches: 93.64,
+      lines: 92.5,
+      functions: 91.74,
     },
   },
   transform: {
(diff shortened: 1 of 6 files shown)


packages/network-of-terms-catalog/catalog/queries/lookup/rtf-personen.rq
```diff
@@ -8,7 +8,7 @@ CONSTRUCT {
         skos:altLabel ?altLabel ;
         skos:scopeNote ?scopeNote ;
         skos:exactMatch ?exactMatch_uri ;
-        rdfs:seeAlso ?seeAlso_uri .
+        rdfs:seeAlso ?uri .
```
ddeboer, line 11: Please simply remove `rdfs:seeAlso` completely: if it’s missing, the website https://termennetwerk.netwerkdigitaalerfgoed.nl will simply direct users to the term’s URI.

ddeboer (changes requested):
Hey @Xentropics, thanks for your PR. Nice improvement on the thesaurus side!

Issue #1742, opened 2026-02-06
Support for corporate authors in the NoT


jolietjakeblues:
**Description**
There is a user request to make corporate authors from the NTA available through the NoT.

At the moment, only personal authors from the NTA are available in the NoT. Corporate authors from the same source do exist in the NTA, but they are not searchable or usable via the NoT.

Examples of corporate authors with PPN that are present in the NTA:

* PPN 175036233
* PPN 075357097
* PPN 355571129

These records can be retrieved, for example, via:
[http://data.bibliotheken.nl/id/thes/p175036233](http://data.bibliotheken.nl/id/thes/p175036233)

**Questions**
It is technically and functionally possible to add corporate authors to the term network.
1. Do we want this?
2. Should this be modeled as a separate service or as an extension of the existing NTA author service?
3. What was the reason corporate authors were not included in the initial implementation?

**Desired outcome**

* Corporate authors from the NTA are searchable and usable via the NoT.
* Clear decision on implementation: extension of the existing service or a separate service.

coret:
I sent KB dataservices (as owner of the [NTA dataset](https://datasetregister.netwerkdigitaalerfgoed.nl/datasets/http://data.bibliotheken.nl/id/dataset/persons)) a message about this issue. Hopefully they can answer some of the questions.

EnnoMeijers:
Thanks for the refinement Djoke! It indeed boils down to the question should the KB Corporatiethesaurus be added to the NoT or not? I think it is up to @jolietjakeblues to decide about this.

Issue #1258, opened 2024-03-11
Add parameters to search queries


rschalkrce:
As discussed in the 11-03-24 NoT meeting, it would be interesting to consider adding (controlled) parameters to search queries. This would allow users to search for specific types of terms, e.g. persons or locations. It could also help integration of NoT URI's into collection management systems by restricting the type of terms that can be used in a certain metadata field. 

Since implementing this is quite a change, both technical and in the management of types we allow, we want to discuss this first.

PR #1747, opened 2026-02-10
Added KB corporation thesaurus

EnnoMeijers:
Could help my self adding this thesaurus after the discussion in #1742 :wink:


packages/catalog/catalog/datasets/kbcorp.jsonld
```diff
@@ -0,0 +1,73 @@
...
+  ],
+  "alternateName": [
+    {
+      "@language": "nl",
+      "@value": "KB Corp"
+    }
+  ],
+  "genre": [
+    {
+      "@id": "https://data.cultureelerfgoed.nl/termennetwerk/onderwerpen/Personen"
```
ddeboer: Is this accurate? Should this perhaps be https://data.cultureelerfgoed.nl/termennetwerk/onderwerpen/organisaties (but note https://github.com/netwerk-digitaal-erfgoed/network-of-terms/issues/1258#issuecomment-3885009594).

packages/catalog/catalog/datasets/kbcorp.jsonld
```diff
@@ -0,0 +1,73 @@
+{
+  "@context": "https://schema.org",
+  "@id": "http://data.bibliotheken.nl/id/dataset/kbcorp",
```
ddeboer: This doesn’t resolve, but http://data.bibliotheken.nl/id/dataset/corps. Should we use the latter?

packages/catalog/catalog/queries/search/kbcorp.rq
```diff
@@ -0,0 +1,34 @@
...
+    ?exactMatch_uri skos:prefLabel ?exactMatch_prefLabel .
+}
+WHERE {
+    {
+        SELECT DISTINCT ?uri WHERE {
+            ?uri schema:mainEntityOfPage/schema:isPartOf <http://data.bibliotheken.nl/id/dataset/corps> ;
+                ?predicate ?label .
+            VALUES ?predicate { rdfs:label schema:name schema:alternateName }
+            ?label <bif:contains> ?virtuosoQuery .
+        }
```
ddeboer: Missing:

```
#LIMIT#
```

packages/catalog/catalog/datasets/kbcorp.jsonld
ddeboer: The decision to identify terminology sources by their distribution was made in #17 (also by you 😁), which I later reverted in #1423. We have kept the distribution URIs for backward compatibility. Just checked the logs, and this is still necessary: many requests still that query (e.g.) AAT by its distribution URIs instead of dataset URIs.

PR #1840, opened 2026-05-01
Use public PiCo URIs in datasets and queries

jolietjakeblues:
- Dataset JSON-LD now uses public URIs (/brontypes, /rollen, /gebeurtenistypes)
- Search and lookup queries rewritten to expose public URIs
- PoolParty URIs retained internally via skos:exactMatch

This aligns termennetwerk output with the public URLs used in the UI.


jolietjakeblues:
Fixes #1754

ddeboer:
https://terms.personsincontext.org/brontypes does not resolve yet, so putting on hold until that is done on the PoolParty side.

We should also keep in mind that this breaks existing URIs, both of the dataset and of the individual concepts. The latter breaks existing references to terms from PiCo.

pwoltjer:
> https://terms.personsincontext.org/brontypes does not resolve yet, so putting on hold until that is done on the PoolParty side.

If you mean https://terms.personsincontext.org/sourcetypes, than I think it does resolve correctly?

packages/catalog/catalog/datasets/picot-brontypen.jsonld
```diff
@@ -1,6 +1,6 @@
 {
   "@context": "https://schema.org",
-  "@id": "https://terms.personsincontext.org/ThesaurusHistorischePersoonsgegevens/523",
+  "@id": "https://terms.personsincontext.org/brontypes",
```
ddeboer, line 3: Should this then become:

```suggestion
  "@id": "https://terms.personsincontext.org/sourcetypes",
```

pwoltjer: Yes, it should be sourcetypes in stead of brontypes

Issue #1907, opened 2026-07-30
Incorrect URI for InC-RUU


ddeboer:
[This query](https://termennetwerk.netwerkdigitaalerfgoed.nl/en?q=In+Copyright+-+Rights-holder(s)+Unlocatable+or+Unidentifiable&datasets=https://data.cultureelerfgoed.nl/rights) returns https://rightsstatements.org/vocab/InC-RUU/1.0/, but that website shows 

> URI for this statement:
> http://rightsstatements.org/vocab/InC-RUU/1.0/

So the Rights thesaurus should return this as HTTP not HTTPS.

The other RightsStatements.org URIs seem to be okay.

jolietjakeblues:
I've fixed this

PR #1908, opened 2026-07-30
fix(catalog): look up Rights terms by their Creative Commons and RightsStatements.org URIs

### LINKED ISSUE #1905: narrower function Right Status terms do not work
Searching for https://termennetwerk.netwerkdigitaalerfgoed.nl/nl?q=creative+commons&datasets=https://data.cultureelerfgoed.nl/rights

To find more fined grained licenses it is useful to select a narrower term like https://termennetwerk.netwerkdigitaalerfgoed.nl/en/lookup?uri=https://creativecommons.org/licenses/by-nc/4.0/ but the look up for this term fails.

ddeboer:
Looking up a term of the Rights Status and Usage Rights thesaurus by its URI failed with `No source found that can provide term with URI …`, so following a narrower term from search results led nowhere.

The thesaurus deliberately reuses Creative Commons and RightsStatements.org URIs as its concept URIs, but the catalog could only route lookups through the single `https://data.cultureelerfgoed.nl/rights/` prefix, which no CC or RS URI starts with.

- Accept multiple `schema:url` terms prefixes per dataset in the catalog loader, for sources whose terms live in more than one URI space. `Dataset.termsPrefixes` was already an array; only the loader collapsed it to one value.
- Declare the `creativecommons.org` and `rightsstatements.org` prefixes on the Rights thesaurus alongside its own.
- Document the convention in the catalog README and pin the routing in a catalog test.
- Raise the catalog-loading hook timeout in `catalog.test.ts` from 20s to 60s. Loading the catalog is CPU-bound (~2s locally) and repeatedly exceeded 20s on CI runners running other tasks in parallel – it fails on master too, e.g. run 30254216562. Unrelated to this change: with and without the `@array` schema change, loading takes 1.6s locally either way.

Both the `http://` and `https://` form of `rightsstatements.org` are declared: the thesaurus publishes eight statements under the canonical `http://rightsstatements.org/vocab/`, but InC-RUU only under `https://rightsstatements.org/vocab/InC-RUU/1.0/` (see #1907). Declaring both keeps lookups working whichever form a client holds, before and after the source data is corrected.

For Creative Commons the source uses `https://` exclusively, so that is the only prefix declared. CC is ambiguous about which form is canonical: the deed page states the canonical URL is `https://`, while its RDF still identifies licenses by `http://` (`<cc:License rdf:about="http://creativecommons.org/licenses/by-nc/4.0/">`) – deliberately so, as CC keeps `http://` inside RDF while migrating human-facing links to https (creativecommons/cc-legal-tools-app#65). Clients holding the `http://` form – common in older heritage metadata – therefore still get no result; whether the thesaurus should use the RDF form is a separate question.

Verified against the live endpoint: `https://creativecommons.org/licenses/by-nc/4.0/` now resolves to the Rights source and returns ‘CC BY-NC’, and `http://rightsstatements.org/vocab/InC/1.0/` returns ‘In Copyright’ / ‘Auteursrechtelijk beschermd’.

Fix #1905

### DIFF
--- a/packages/catalog/README.md
+++ b/packages/catalog/README.md
@@ -33,7 +33,9 @@ additions specific to the Network of Terms:
   resources. This prefix is needed when clients look up terms by their URI in the Network of Terms: the Network then has
   to know which source to consult to retrieve the term. If the dataset is a subset of a broader dataset in the same URI
   space (e.g. ‘Wikidata: persons’ is a subset of ‘Wikidata: all entities’), omit `schema:url`: only the
-  broadest dataset declares the prefix, so that lookups resolve to it rather than to an arbitrary subset;
+  broadest dataset declares the prefix, so that lookups resolve to it rather than to an arbitrary subset. A dataset
+  that holds terms in more than one URI space declares `schema:url` once per prefix, for example the Rights thesaurus,
+  which reuses the canonical Creative Commons and RightsStatements.org URIs alongside its own;
 - `schema:inLanguage` is a required property;
 - `schema:genre` is a required property, with values restricted to the list of [Termennetwerk onderwerpen](https://data.cultureelerfgoed.nl/termennetwerk/onderwerpen.html);
 - `schema:mainEntityOfPage` is a required property;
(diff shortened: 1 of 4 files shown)


jolietjakeblues:
fixed

PR #1909, opened 2026-07-31
fix: updated ABR to ABR 2

jolietjakeblues:
RCE heeft het Archeologisch Basis Register geupdate. De uri's hebben nu een /2/ .

### DIFF
--- a/packages/catalog/catalog/datasets/abr.jsonld
+++ b/packages/catalog/catalog/datasets/abr.jsonld
@@ -1,6 +1,6 @@
 {
   "@context": "https://schema.org",
-  "@id": "https://data.cultureelerfgoed.nl/term/id/abr/",
+  "@id": "https://data.cultureelerfgoed.nl/term/id/abr",
   "@type": "Dataset",
   "name": [
     {
--- a/packages/catalog/catalog/datasets/abr.jsonld
+++ b/packages/catalog/catalog/datasets/abr.jsonld
@@ -1,6 +1,6 @@
 {
   "@context": "https://schema.org",
-  "@id": "https://data.cultureelerfgoed.nl/term/id/abr/2/",
+  "@id": "https://data.cultureelerfgoed.nl/term/id/abr/",
   "@type": "Dataset",
   "name": [
     {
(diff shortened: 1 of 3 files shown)


ddeboer:
But changing the URIs will break existing users that have already linked v1 URIs: they can no longer look those up in the Network of Terms.

Why did the URIs change? That’s rather bad practice. URIs must be persistent.

jolietjakeblues:
David,

The v1 URIs have not disappeared and they still resolve. We deliberately keep the deprecated version online so existing links remain valid.

The reason the URIs changed is that the ABR is a versioned vocabulary. Every 3 to 4 years, RCE and SIKB publish and formally adopt a new version of the Archaeological Base Register. Once a new version is released, the previous one is marked as deprecated.

We intentionally include the version number in the URI because each published version is a fixed snapshot. This allows users to reference the exact version that was used in a dataset or publication, while also making newer versions available.

We encourage users to migrate to the v2 URIs, but we do not break existing references. The v1 URIs continue to resolve and remain available for backwards compatibility.

Also, for concepts that exist in both versions, the UUID is identical. Only the version segment in the URI changes. This makes it straightforward to identify the corresponding concept across versions while still distinguishing between published releases.

We are also working on adding explicit provenance metadata to concepts in versions >1. Once that is in place, users will be able to trace the lineage between concept versions more easily. For now, version-specific URIs combined with stable UUIDs provide a pragmatic solution: the UUID identifies the concept across releases, while the URI identifies the specific published version.

ddeboer:
Thanks for the explanation @jolietjakeblues! That makes sense, and I’m glad the v1 URIs stay resolvable. Lookups for v1 terms _will_ continue to work this way. 

The real remaining problem is the change of the dataset URI (`@id`) from `.../term/id/abr` to `../term/id/abr/2`. In the Network of Terms that is the _source identifier_: clients (including collection management systems) pass it through to indicate that they want to search ABR.

Could you change that one line back and keep the rest? Searches will then continue to work, now returning v2 items.

jolietjakeblues:
Will do

ddeboer:
Thanks!

PR #1924, opened 2026-08-03
feat(catalog): add Brandweerthesaurus

jolietjakeblues:
Nieuwe versie.
Fout in vorige pull request.
Concepten resolven


packages/catalog/catalog/datasets/korpora.jsonld
```diff
@@ -0,0 +1,77 @@
+{
+  "@context": "https://schema.org",
+  "@id": "https://www.korpora.info/data/rdf",
```
ddeboer: This URI doesn’t resolve. Is that a problem? Do we have a better URI?

jolietjakeblues: It's what they want as uri. Maybe we can talk about it tomorrow a bit. There is another problem. You've probably already guessed it. The URI identifies the concept itself, not a web page displaying the preferred term. The concept URI in the thesaurus ends with /rdf/ followed by a UUID. Because this the server immediately performs content negotiation. If the client requests RDF, the server returns an RDF representation, which triggers a download in a web browser. As a result, users never see the concept as a human-readable web page. Instead, they receive the RDF representation of the concept.

